PR #7, opened 2022-07-30
Add README

NotNite:


philpax:
lgtm but think it needs mention of DalamudPackager and how to set it up too

NotNite:
I wrote this in the context of plugin developers already knowledgeable with submitting to DalamudPlugins, not sure it's worth it. When we actually deploy Plogon, I'll do it

philpax:
We have a few people not using DP right now, figure they could benefit from a migration guide. Should just be able to paste in the bit from the DIP about the properties I think

PR #19, opened 2022-07-31
Manifest validation on PR

nonowazu:
Like it says on the tin, this sets up a gh workflow to validate any `.toml` files that get added to the `stable` or `testing` directories. I'm very open to changing paths (for example, moving/renaming pyckager) to keep things a bit cleaner in the root of the repo.

To change how the toml file is validated, update the [pydantic](https://pydantic-docs.helpmanual.io/usage/models/) models in `pyckager/models.py`.

One caveat is that this will break on any *new* toml file added that uses github IDs rather than strings (see the validator in the model for more information; I'm also open to changing this as well).

### DIFF
--- a/.github/workflows/validate_manifest.yml
+++ b/.github/workflows/validate_manifest.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
 
     - name: Check for manifest changes
       id: changed_manifests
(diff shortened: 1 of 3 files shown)


.github/workflows/validate_manifest.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    - name: Setup Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+
+    - name: Install python deps
+      run: python -mpip install pydantic==1.9.1 toml==0.10.2
+
+      # this does technically break on a filename with spaces, but nobody would do that, right? ...right?
```
NotNite, line 32: InternalName shouldn't have spaces so this shouldn't be an issue

nonowazu: I'd rather document it somewhere than let it bite someone later; that's all.

goaaats (approved):

PR #25, opened 2022-07-31
Add version checking to manifest validation

nonowazu:
What is says on the tin, but the short of it is:

* Based on some discussion in-channel, semver actually doesn't respect 4-segment versions (`w.x.y.z`), so I left us with a much simpler `^(\d+\.)+\d+$` regex (which I validate to `w` version all the way to `w.x.y.z` in the validator function)
* (Doesn't change the `validate` subcommand in the cli) parses the version as type `tuple[int]` – not useful at the moment, but could be nice for csproj comparison later with the same tool

### DIFF
--- a/pyckager/models.py
+++ b/pyckager/models.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel, validator
 
 
 NUMBERS_ONLY = compile(r'^\d+$')
+VERSION = compile(r'^(\d+\.)+\d+$')
 
 class Plugin(BaseModel):
     repository: str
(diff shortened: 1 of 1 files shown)


goaaats:
It's still optional to have here, right?

nonowazu:
Yes, the annotation in the model marks it as optional:

```python
version: Optional[str]
```

When pydantic instantiates the model, it just makes the version `None`:

```python
>>> from pyckager.models import Plugin, Manifest
>>> Manifest(plugin=Plugin(repository='foo', commit='foo', owners=['nonowazu']))
Manifest(plugin=Plugin(repository='foo', commit='foo', owners=['nonowazu'], project_path=None, changelog=None, version=None))
>>> Manifest(plugin=Plugin(repository='foo', commit='foo', owners=['nonowazu'], version='1.0'))
Manifest(plugin=Plugin(repository='foo', commit='foo', owners=['nonowazu'], project_path=None, changelog=None, version=(1, 0)))
```

PR #35, opened 2022-08-02
CrossUp 0.3.3.0

ItsBexy:
- Made changes so that the Expanded Hold Bars should properly match the visibility state of the main Cross Hotbar in most cases.
 - Added feature to auto-fade the Cross Hotbar to a custom transparency outside of combat.
 - Updated to net6.0 (so far so good)


NotNite:
That's the wrong testing folder for net6.

PR #67, opened 2022-08-08
[Testing] Allagan Tools v1.1.0.7

Critical-Impact:
Update for net6, also this is just "Inventory Tools" renamed, the assembly name has not changed, just the name in the json and manifest. Rename was required as it not longer just handles inventory.
No idea if this'll work either, quite a few moving parts so fingers crossed I guess?


NotNite:
What will happen with two plugins with the same assembly name but diff name in pluginmaster?

Critical-Impact:
> What will happen with two plugins with the same assembly name but diff name in pluginmaster?

I did ask the question before I renamed, apparently if I leave the assembly name alone and just rename it should be fine, though I'm sure someone might have a more definitive answer. From my testing(at least with dev plugins), a name change doesn't register straight up and only seems to show up once you've restarted though this might different for live/testing plugins.

reiichi001:
bleatbot, approve

reiichi001 (approved):

PR #87, opened 2022-08-17
Updating DelvUI

Tischel:


Tischel:
Not sure if the warning from the bot is important or not.
The reason why the version is not changing is because this will be the version that we'll release on patch 6.2, in the meantime we're just updating it with all the changes that are being pushed to the public dalamud version.

goaaats:
it's just a warning, in case people forget to bump it

goaaats (approved):

PR #106, opened 2022-08-21
Spaces in paths not work for manifest validation

nonowazu:
Should work for spaces in paths now, with the notable point that `,` is now no longer allowed in paths instead of spaces.


philpax:
This looks reasonable to me, so let's give it a shot

PR #109, opened 2022-08-21
submit EngageTimer 2.2.2.2 to dotnet6 testing

xorus:
Did I do it correctly?


philpax:
bleatbot, approve

NotNite:
Seems like your lockfile isn't complete - missing valuetuple. We saw this earlier with someone else, I think the fix was just run dotnet restore again

philpax:
Should be good to go once you've got the lockfile sorted out

xorus:
I'm couldn't find any usage of ValueTuple anywhere so this might be a depedency thing (or my find in files does not work). Either way I'll add it to the package list.

philpax:
Shouldn't need to add `ValueTuple` directly, this looks like a framework mismatch. Try deleting the lockfile and running `dotnet restore` so that it's forced to regenerate it for .NET6

xorus:
Do I remove it ?

philpax:
Doesn't matter either way, but you're not using it directly, so I would.

xorus:
maybe it's beacause NAudio has a dependency on winforms, even though I don't use that

philpax:
@goaaats any idea what's going on here?

xorus:
For ValueTuple :
EmbedIO requires Unosquare.Swan.List that requires ValueTuple from dotnet framework 461 - I guess nuget cannot resolve this.
So I added back the ValueTuple nuget (it does not let me restore wihout now), and also removed the NAudio part (this was already planned for later, it's been marked as "legacy - contact me if you use it" for a while now).
Let's try this now

philpax:
yay!

goaaats:
Realistically the solution is "build in windows containers" but I'm not sure if that's feasible in GitHub actions, never tried

xorus:
> Realistically the solution is "build in windows containers" but I'm not sure if that's feasible in GitHub actions, never tried

I was able to completly remove the NAudio lib completly but I don't think other plugins that use it can do that. (assuming it was the problem in the fist place)

xorus:
@philpax a heads up if you meant to auto merge, it got blocked

reiichi001 (approved):

PR #112, opened 2022-08-21
The rest of the Anna plugins except XIVChat

ascclemens:
XIVChat is complicated to do from my laptop.


goaaats (dismissed):

goaaats:
Seems like NoSol has libs that are still on net5, can you upgrade them? If not, I have to tweak the dep resolution stuff a little

ascclemens:
Oh oops, included XIVChat.

ascclemens:
Peeping Tom looks like issues with spaces again. This one is unreasonable to ask me to change; this is internal hierarchy of my code. Your system should be able to handle that.

goaaats:
https://github.com/goatcorp/DalamudPluginsD17/runs/7941471245?check_suite_focus=true#step:9:317
Do you need this for sound?

ascclemens:
I remember that I couldn't get rid of it back when I updated NAudio, despite my trying. I think it's probably for listing the audio devices? Give me a few and I'll see if it's required.

goaaats:
I've been trying to come up with a solution for "we can't afford windows infrastructure" and coming up dry so if there's any way you can do without, that'd be good. @xorus how'd you solve it?

ascclemens:
As far as I know, there's no way around it. The entire `NAudio.Wave` namespace is in there, and that's how literally all of the sound stuff works. Unless there's some alternate API that lets me do the same things as `NAudio.Wave` does, which would be amazing.

goaaats:
Looking at the NAudio code, the only reason the winforms stuff is in there in the first place is that it needs a win32 window to play sound through, so we might add an API for sound playback to Dalamud that mirrors NAudio.Wave and can also hook up to the XIV sound mixer.

ascclemens:
That would work for me, too. I also need a way to enumerate sound devices and decide what sound device to play through, as well as the volume the sound will play at. Not sure if enumerating the devices is also in `NAudio.Wave` or not.

xorus:
@goaaats I removed the entière audio lib :|.

It was only there as a legacy thing because back then I didn't have a way to play the in game timer sound.

NAudio just does not work well on Linux too. I've had reports of crashes on lonux that were caused by the lib. 

It'd be indeed neat to have an API to play sound.

ascclemens:
So do I remove Peeping Tom from this PR? Can't use NAudio anymore? Or I could provide my own assemblies for that specific plugin if the CI can't do it.

goaaats:
We're working on it, should have something before it matters

ascclemens:
For future reference, `NAudio.Core` and `NAudio.Wasapi` using the `DirectSoundOut` interface works without a dependency on WinForms. It also removes the half-second hang the first time a sound is played (presumably creating a window or something).

goaaats:
We were planning to offer an option to build with a wine container, but I guess we can just keep that in the back of our mind for now

goaaats (approved):

PR #121, opened 2022-08-21
Add PingPlugin and SoundSetter

karashiiro:
Expecting CI failures, will feex


karashiiro:
bleatbot, approve

philpax:
Given TTT will be broken until we figure out how to tackle Windows builds, maybe remove it from this PR?

karashiiro:
I'm working on it, I use Windows Forms for file dialogs and NAudio for sound. I'm replacing the Windows Forms dialogs with imgui ones, and the .NET 6 version of NAudio doesn't necessarily require the Windows platform target. The platform-agnostic version of NAudio doesn't have the sound file interface I was using, but I might be able to get it working with a different sound file interface.

karashiiro:
k nevermind

philpax (approved):

PR #169, opened 2022-08-24
MP Tick Bar .net6

marconsou:
- MP Tick Bar .net6


Aireil:
You should PR to stable/ or testing/live so it actually shows on the installer, testing/net6 does nothing.

marconsou:
Yeah, i was thinking right now about it.

PR #258, opened 2022-08-26
[Testing] Update KikoGuide to 1.4.1.0

BitsOfAByte:
Fixes a plugin crash for first time users when a timestamp was saved to the configuration file.

If you are experiencing crashes when the plugin launches, you will need to delete your configuration which can be found at `AppData/Roaming/XIVLauncher/pluginConfigs/KikoGuide.json` and restart


BitsOfAByte:
Need to fix something with this first, just realised

reiichi001 (approved):

PR #265, opened 2022-08-26
[nofranz] EnemyListHP: Added former name into plugin's name as well

Eternita-S:
I'm just adding former name because I get quite often questions about people remembering it as EnmityHp and trying to search for the plugin in installer but unable to find it. Will just keep it for some time and will remove it with some further update.


Eternita-S:
same version, since there's no code changes and no need to enforce people to update (obviously people who already have it installed know the new name)

philpax (approved):

PR #279, opened 2022-08-27
Update Chat 2 and ExtraChat

ascclemens:
This update allows users of both Chat 2 and ExtraChat to filter ExtraChat messages in their Chat 2 tabs!

(Note that this only applies to messages received after updating ExtraChat.)


ascclemens:
Heh, ExtraChat also shows the server diff. Maybe scope the diff to the folder specified in manifest?

goaaats (approved):

PR #282, opened 2022-08-27
[Testing] Add WhichPatchWasThat 1.0

Kouzukii:
Which Patch Was That? lets you quickly figure out the patch your items were released in.

Never carry around old useless crafting materials again!

image1


philpax:

goaaats (approved):

PR #300, opened 2022-08-28
Update NoKillPlugin

Bluefissure:
It's not actively useful rn but in case the lobby blows up again

### DIFF
--- a/testing/live/NoKillPlugin/manifest.toml
+++ b/testing/live/NoKillPlugin/manifest.toml
@@ -1,0 +1,5 @@
+[plugin]
+repository = "https://github.com/Bluefissure/NoKillPlugin.git"
+commit = "a361208fa5d732b5bcd31e34338c998fd1125a56"
+owners = ["Bluefissure"]
+project_path = "NoKillPlugin"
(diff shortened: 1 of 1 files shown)


testing/live/NoKillPlugin/manifest.toml
```diff
@@ -0,0 +1,5 @@
+[plugin]
+repository = "https://github.com/Bluefissure/NoKillPlugin.git"
+commit = "a361208fa5d732b5bcd31e34338c998fd1125a56"
+owners = ["Bluefissure"]
+project_path = "NoKillPlugin"
```
karashiiro, line 5: You can remove this line, this refers to where your solution is stored.

NotNite: I don't think you can? I'm pretty sure it's where the .csproj is stored, not the solution.

karashiiro: Oh? That's how it seemed to work when I added PingPlugin and TTT. TTT has the solution file in `src`, so I had to adjust this field. PingPlugin didn't need the field at all.

karashiiro:
https://github.com/Bluefissure/NoKillPlugin/blob/master/NoKillPlugin/NoKillPlugin.csproj#L36-L38

This time it's this messing up the workflow. DalamudPackager should do this already anyways.

Bluefissure:
> https://github.com/Bluefissure/NoKillPlugin/blob/master/NoKillPlugin/NoKillPlugin.csproj#L36-L38
> 
> This time it's this messing up the workflow. DalamudPackager should do this already anyways.

Yeah, some of my plugins have weird legacy build flows that mess up the CI.

karashiiro (approved):

PR #305, opened 2022-08-29
[Testing] add abilityants to testing

attickdoor:
abilityants is a plugin that lets the user select abilities (not weaponskills or spells [except for the weird ones]) to get the combo ants around them when off CD. They can also offset by an arbitrary number of milliseconds so that the ants show up earlier.


philpax:

lmcintyre:
This won't impact your plugins' approval, but just so you know, it behaves a little odd on actions with multiple charges. I used Soul Slice for testing and it only added the ants when it had 2 charges.

attickdoor:
The behavior seems correct on Soul Slice. If "ants only on the final charge" is selected, the ants only show up at the correct offset for charge 2. If it's not selected, then the ants will be active effectively starting at the offset, but for the first charge. The difference is that the former means "you're about to cap charges" and the latter means "you have a charge you can use".

lmcintyre:
> The behavior seems correct on Soul Slice. If "ants only on the final charge" is selected, the ants only show up at the correct offset for charge 2. If it's not selected, then the ants will be active effectively starting at the offset, but for the first charge. The difference is that the former means "you're about to cap charges" and the latter means "you have a charge you can use".

Yeah so I'm simply incapable of reading and you can completely ignore what I said.

lmcintyre (approved):

PR #307, opened 2022-08-29
Update Tourist

ascclemens:
It's fixed now, okay?


ascclemens:
Surely not?

```xml
<Reference Include="Lumina">
    <HintPath>$(DalamudLibPath)\Lumina.dll</HintPath>
    <Private>False</Private>
</Reference>
```

PR #320, opened 2022-08-30
Update DeviceChangeFix

Jessidhia:
Removes the config window, replaced by the new disable switch in the plugin list.


Jessidhia:
Looks like the port to SignatureHelper was done incorrectly and the plugin is causing crashes after unload, update retracted until it's fixed.

PR #328, opened 2022-08-30
DailyDuty v3.0.1.4

MidoriKami:
- `Timers` Added new feature to hide timers for completed tasks (Thanks for the suggestion! xywzel)
- `Wondrous Tails` Fixed Duty start/end notifications showing when you don't have a wondrous tails book

- `Configuration` Added fallback in case configuration migration fails
***If a configuration file fails to migrate, your settings will be reset***
***This only applies to users of DailyDuty before patch 6.2***

image

### DIFF
--- a/stable/DailyDuty/manifest.toml
+++ b/stable/DailyDuty/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/DailyDuty.git"
-commit = "5229df85d6b1284bd32c5ab75381563428efb9cc"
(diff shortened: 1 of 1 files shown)


philpax:
Manual diff because GitHub Actions is being a bit slow today: https://github.com/MidoriKami/DailyDuty/compare/5229df85d6b1284bd32c5ab75381563428efb9cc..1fc9bb4c45ff907c5d067d19f1073ccb9bb081b0

philpax (approved):

PR #362, opened 2022-09-02
chore: remove stg overrides

Aireil:
Once again Aireil coming up with an untested PR, what could go wrong?
I have no idea how to remove this, but if it is not like this, consider this as an issue instead 😃.

### DIFF
--- a/overrides.toml
+++ b/overrides.toml
@@ -1,3 +1,1 @@
 [channel_tracks]
-testing-live = "stg"
-stable = "stg"


philpax:
is this still relevant?

Aireil:
Yes, although it's mainly there so I can say "See, I told you!" when it eventually breaks something.

philpax:
I'll reopen when something breaks so that you can get your glory

PR #397, opened 2022-09-04
add discord user id fields to plugin model

nonowazu:
Probably need to discuss what the field will be named, but eh, it's a 2s change.

### DIFF
--- a/pyckager/models.py
+++ b/pyckager/models.py
@@ -11,6 +11,7 @@ class Plugin(BaseModel):
     repository: str
     commit: str
     owners: list[str]
+    discord_ids: Optional[list[int]]
     project_path: Optional[str]
     changelog: Optional[str]
     version: Optional[str]


nonowazu:
Also, it looks like bleatbot will always refuse to allow me in because I'm never submitting plugins - only changes to manifest validation?

philpax:
bleatbot, approve

philpax:
hopefully that frees it from needing approval again? if not I'll open an issue on https://github.com/goatcorp/worker-plugindispatch

anyway looks good, just holding off on the merge because still deciding if we want it to be part of the `owners` field or separate from it. e.g. the difference between these three:
```typescript
owners: {github: string, discord: string},
```
```typescript
owners: [string, string][],
```
```typescript
owners: string[],
discord_ids: number[]
```

please maintain a holding pattern until a decision is made 🙏

nonowazu:
I don't think changing owners from a list to an object is a good idea. If the purpose is to act as a notification mechanism, perhaps:

```
notifications: {
    discord: int,
    email: str, # as an example
```

or a list of those object types.

philpax:
Yeah, I was thinking it'd be either format: `string[] | {discord, email}[]`, so that there's a link between the owner and their Discord username, without breaking existing manifests.

Notifications aren't the worst idea. Gotta think about it some more.

philpax:
I'm going to close this until we figure out more decisively what we want to do here - will reopen once we do! Thanks for the help ❤️

PR #406, opened 2022-09-05
Add Compass 1.26.1

fitzchivalrik:
- chore: Update to API 7


philpax:
Did you mean to leave that target system pointer text in?

fitzchivalrik:
It is part of `PluginDebug` which is only added when compiled in 'Debug' configuration.
Aka, I did not clean up my debug stuff completely as it is not included in Release mode anyway x).

PR #411, opened 2022-09-05
ChillFrames v1.5.0.8

MidoriKami:
Probably finally actually fix blacklist behavior
Fixed version text clipping other ui elements


philpax:
feel like there's gotta be a simpler way to express that logic but hey if it works it works

philpax (approved):

PR #415, opened 2022-09-06
[Stable] Fix broken message translation

BitsOfAByte:
- Fixes a bug where after a user has installed the plugin, some messages will not be properly localized and will not show the proper placeholders.
- Minor improvements on the server-side.


BitsOfAByte:
oops :^)

BitsOfAByte:
ok ill fix this some other time nvm im too tired

PR #419, opened 2022-09-06
[Testing] NoTankYou v5.0.0.2

MidoriKami:
Total internal rework, everything should function the same, but better.

All configurations are now per-character, your old config will be copied to each of your characters.

image

### DIFF
--- a/testing/live/NoTankYou/manifest.toml
+++ b/testing/live/NoTankYou/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/NoTankYou.git"
-commit = "297e553753f4b6d0acdaf8f6ad967a5f7e997121"
+commit = "672a7b1adc24c8be4c02d3e3e505b7c743ead270"
 owners = ["MidoriKami"]
 project_path = "NoTankYou"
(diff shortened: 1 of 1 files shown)


MidoriKami:
There were no added nor removed features in this version, just system reworks.
User configs won't actually be wiped, they will just be ignored. I'm switching from using NoTankYou.json to using LocalContentID.json inside the `pluginConfig\NoTankYou` folder. As such, a user could revert from testing to get their configuration back.

The new design makes it very easy to add new modules, so in the coming days I will be adding some more features.

philpax:
This is a pretty big diff so won't be able to review it for a bit - would suggest using that time to figure out a migration story

lmcintyre:
I can't say I know much about what there is to configure for NoTankYou; however; I would recommend migrating as well. Even if it's as simple as "copy old config to all characters" kind of thing. Otherwise people will generally be very upset. If you are fine with that, we can't hold you back from your decision though.

MidoriKami:
I've already been working on it.

MidoriKami:
There done. That should make everyone happy.

MidoriKami:
Added PartyListOverlay preview mode, there will be no more changes for this PR.

lmcintyre (approved):

PR #442, opened 2022-09-08
[Testing] NoTankYou v5.0.0.5

MidoriKami:
Fix race condition that breaks all dalamud font colors. (Thanks Chalkos)

A race condition is when one part of code runs faster or slower than another part of code.
For this to have happened, the function had to have been executed within a 1millisecond window and completed at the start of another 1millisecond window. This change removes the race condition. (Fixes it)


MidoriKami:
Tossed in a scaling fix that was recently reported for DailyDuty, since it uses the same code it had the same issue.

philpax (approved):

PR #449, opened 2022-09-09
Update Orange Guidance Tomestone to 1.6.1

ascclemens:
- Added new glyph.
- Added option to hide signs during gpose and cutscenes.
- Added ban list interface to settings.
- Added warning for low-score messages.


philpax (approved):
this mf said
> (that aren't the from the user)

PR #457, opened 2022-09-10
[Testing] MiniMappingway

jaycewhite:
Now autodetects if minimap is locked. Added a temporary warning about fc marker functionality.


jaycewhite:
I accidentally left a value I changed for debugging set

reiichi001 (approved):

PR #466, opened 2022-09-11
NoTankYou v5.0.0.7

MidoriKami:
Big internal rework!

User interface is simplified significantly, less words to not read!

All settings and window positions are now ***per character***, Now you can enable and disable settings and position the warning banner however you want for each of your characters.

You can now mix-and-match your warning display modes, each module has a checkbox to enable Party Frame and Banner overlays. 
Want food notices to be subtle in the party frame? 
Want a big warning when you forgot your faerie?
Now you can!

Added Sample Modes for Party Frame Overlay, and Banner Overlay. You can now see what the warnings will look like and more easily reposition the Banner Overlay. 
Sample Mode will ignore any pending actual warnings, and will override other things that would otherwise prevent warnings from showing such as Hide in Sanctuaries.
No more messing with a bunch of settings just to see what the warnings will look like.

And lastly, major font/text improvements to the Banner Overlay (Thanks Kizer!)
The text should display very nicely without blurring (Except when you scale it up massively of course)


MidoriKami:
This release is simply moving the Testing version to Stable.
There have been Zero code changes between Testing and Release. 
I did not increase the version number for this release.

philpax (approved):

PR #473, opened 2022-09-12
[Testing] DailyDuty v3.0.3.2

MidoriKami:
`Raids (Normal)` Use DutyFinder Refresh event to refresh only when needed
`Raids (Alliance)` Use DutyFinder Refresh event to refresh only when needed
`Raids (Normal)` Filter drops on `Miscellany` category


MidoriKami:
@philpax There, a small diff this time.

philpax (approved):

PR #474, opened 2022-09-13
WaymarkPresetPlugin 1.4.3.2

PunishedPineapple:
- Disabled access to the preset editor following SE's complaints about out of bounds (OOB) waymarks.  This will probably be a temporary restriction if further safegaurds can easily be implemented.
- This was done to make it a bit harder for people to shoot themselves in the foot.  It's still possible to place OOB waymarks in the same way the game allows you to do so, so use your brain.
- Any presets you import from outside of the game, I'd personally advise checking on the map view for OOB waymarks before using.


PunishedPineapple:
Please lmk if you'd like for me to restrict things further than this.

PR #478, opened 2022-09-14
NoTankYou v5.0.1.0

MidoriKami:
New Feature! Disable in Sanctuary is now available **for each module**

You can now disable and enable the disable in sanctuary option individually in each module.
Now you can be warned about Spiritbond and Free Company buffs while in town!

(Thanks anonymous feedback person)


MidoriKami:
Added new "Solo Mode" display mode feature to limit warnings to player only.
Fixed bug that prevented window titles from updating with correct character name.

PR #509, opened 2022-09-18
[Testing] DailyDuty v3.0.5.0

MidoriKami:
New Module! `Unreal Trials`

Track your weekly completion and re-completion of the current unreal trial.

image
image


MidoriKami:
This update does include a "SendChat" function taken from XivCommon. In my implementation of it, a `/` is forced in-front of any message sent. No user-input data will ever be sent to chat via SendChat by DailyDuty. SendChat will only be used to send client-side commands.

philpax (approved):

PR #519, opened 2022-09-19
[Testing] MiniMappingway - Big Refactor

jaycewhite:
Big refactor to increase performance and enable other plugins to integrate.


jaycewhite:
I need to fix some stuff, I don't want this deployed lol

PR #535, opened 2022-09-22
Update MapLinker to v1.1.7.0

Bluefissure:
Activate game & Export maplinks


Bluefissure:
Not sure whether automatically placing the flag when detected maplinks in chat exceeds the no-auto boundary, if it does I can remove that.

reiichi001 (approved):

PR #537, opened 2022-09-22
[Testing] KitchenSync v1.0.0.2

MidoriKami:
Add new plugin, KitchenSync!

KitchenSync allows you to change the transparency value of any skills that aren't available due to being level sync'd.

This is just an initial version, many more features are planned! Let me know if you have any ideas or requests!
(Note: Any suggestions that may provide an advantage in combat will be rejected)

image


MidoriKami:
I managed to break something, so this is a resubmission with some fixes I found while testing it myself.

philpax (approved):

PR #539, opened 2022-09-23
[Testing] Slice is Right 5.1

carvelli:
New plugin - adds overlays to the Slice of Right GATE to see where you need to go. 

The overlays should render roughly around the same time as the bamboo cut animation happens.

Example:

grafik


reiichi001:
This plugin violates our plugin guidelines. It creates a clear and direct advantage over players not using plugins by showing visible AOE markers for mechanics that aren't normally telegraphed to the client.

PR #554, opened 2022-09-26
[Testing] MiniMappingway - Massive Refactor

jaycewhite:
Massive refactor. Everyone option added. New settings screen. Border option available for markers now. Able to set marker styles per group. Added possibility for other plugins to integrate.


philpax (approved):
Nice work! It's much cleaner now 🙂

PR #562, opened 2022-09-27
[Testing] KitchenSync v1.0.0.7

MidoriKami:
Minor system improvements

Added `Apply to Macros` option that is on by default.

Reworked how Extended Crosshotbars work, should work well for controller users now.

image


philpax (approved):
*removing* helpers? madness

PR #592, opened 2022-10-04
AutoMammet 1.2.0.1

Rietty:
Simply updating a bit of the wording at specific locations to indicate where data is exported to, in order to be more clear based off recent feedback.


philpax:
Seems reasonable, but you might want to also add this to the point of use so they know without checking the description

philpax (approved):

Rietty:
I will do that in the future as well, I figured people would read the description when they installed it.

PR #598, opened 2022-10-07
KitchenSync v1.0.2.1

MidoriKami:
Added `Apply to not yet unlocked` option to apply transparency to skills you don't have unlocked yet
Added `Potato Mode` to slow down the rate that hotbars are checked. (Shouldn't ever need this but it's there if you do)


MidoriKami:
This PR is just a "move from testing/live -> stable", there were no code changes.

philpax (approved):

PR #606, opened 2022-10-09
[Testing] Add RoleplayProfiles v0.1.0.0

Maia-Everett:
A plugin for Europe region roleplayers, displaying profiles from Chaos Archives (https://chaosarchives.org) for characters with the RP flag on.

Future releases will add the ability to edit your own character profile in-game, and later, possibly, extension to other regions.

### DIFF
--- a/testing/live/RoleplayProfiles/manifest.toml
+++ b/testing/live/RoleplayProfiles/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Maia-Everett/dalamud-roleplay-profiles.git"
 commit = "b0685cbd7767884bf6be7b143b52d03fcdf0c330"
-owners = ["Vielle Janlenoux"]
+owners = ["Maia-Everett"]
 project_path = "RoleplayProfiles"
 changelog = "Initial test release (displaying profiles from Chaos Archives)"
--- a/testing/live/RoleplayProfiles/manifest.toml
+++ b/testing/live/RoleplayProfiles/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Maia-Everett/dalamud-roleplay-profiles.git"
-commit = "0ed6d69904bf2ad140fd767c53f862b97f9ac701"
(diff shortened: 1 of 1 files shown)


Maia-Everett:
Screenshot

philpax:

Maia-Everett:
Here's the documentation for the API used by this plugin:

https://chaosarchives.org/api/rpp/swagger/#/

As an example, here's my character profile, as retrieved via this API:

https://chaosarchives.org/api/rpp/profile/Omega/Vielle%20Janlenoux

Maia-Everett:
v0.1.0.0 should be the final state of the plugin, feature-wise. Only bugfixes will follow.

This version adds listening to character profile updates over WebSocket, so that profiles displayed in-game are updated instantly.

reiichi001:
Hello! During review, I noticed a minor issue on the initial config window. It looks like the right side of the window has some text being clipped. I would recommend making the window resizable for a quick fix, or to recheck it for Dalamud font scaling so that it can be adjusted as needed without text getting clipped from view.

Maia-Everett:
> Hello! During review, I noticed a minor issue on the initial config window. It looks like the right side of the window has some text being clipped. I would recommend making the window resizable for a quick fix, or to recheck it for Dalamud font scaling so that it can be adjusted as needed without text getting clipped from view.

Hmm, could you attach a screenshot? It looked fine to me, though I use a 4k monitor with 200% font scaling.

I can make it resizable, or I can reproduce it and then adjust size as necessary.

reiichi001:
Here's what I mean. This is on pretty default-ish settings, but I do have the axis font enabled. There's some clipping on the right size of the box, and especially for an email, you can't see some of the text, which would make validating it's been typed correctly a bit difficult.

image
image

Maia-Everett:
This is extremely weird. The config window on your screenshots is about half the width and height that it should be.

I'm not sure why this is happening. I've tested with different font sizes and couldn't reproduce the issue.

As a workaround, I've made the config window resizable in version 0.1.0.1.

philpax (approved):

philpax:
Sorry, should've gotten back to you earlier! Thanks for submitting another plugin :)

PR #627, opened 2022-10-18
DailyDuty v3.0.7.0

MidoriKami:
New modules!

This update adds grand company supply and provisioning!

These modules update while the Grand Company supply and provisioning window is open. 
(It will not update from the in-game timers submenu).

Add `Grand Company (Supply)`
Add `Grand Company (Provision)`

YoloMouse_2iWZsZ7I4C
image


MidoriKami:
This PR is only a move from Testing to Stable, there were no changes.

PR #641, opened 2022-10-22
[Testing] Update GoodFriend to v1.4.0.0

BitsOfAByte:
A complete overhaul of GoodFriend, making major improvements. Thank you to KazWolfe for helping me resolve some problems that were blocking this release.

There may be some bugs in this release which is why it is being sent to testing first. Please note that when using the testing build you will be unable to share events with users on the stable build.

**__Features__**
- A new, redesigned UI that is way clearer.
- Improvements to 3rd party instance support (instances can now set the status button, "Instance Host" donation button on their side.)
- Rate limit support, preventing the plugin from spamming an instance when rate limited.
- A full event log showing requests to the API and other things, helping make all interaction with the API easier to audit; also making it easier to diagnose any errors.
- The API instance can now ask the plugin to update your API-Url in the case of, for example, a domain move. You will always be prompted to accept this change.
- Better clarity on all tooltips, making it way easier for users of all tech understandings to configure the plugin. 
- Performance and usability improvements. 

**__Bug Fixes__**
- Resolved an issue with "Ignore FC Members" wherein friends with the same FC tag on other homeworlds were also ignored.
- Many minor bugs under the hood.

### DIFF
--- a/testing/live/GoodFriend/manifest.toml
+++ b/testing/live/GoodFriend/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/BitsOfAByte/GoodFriend.git"
-commit = "0f493a7231b7c6fac6c242442d39472c7127e984"
+commit = "22f55179b43de801aad2b94cc3193300a2579ed1"
 owners = [
     "BitsOfAByte",
 ]


BitsOfAByte:
(lint-manifest will not run due to the whole GHA problem, if it's needed then someone will have to trigger it by pushing to this branch)

karashiiro (approved):

karashiiro:
@goaaats will need to manually merge this if the lint action won't run.

PR #649, opened 2022-10-24
[Testing] Island Workshop Solver - update manifest

rosella500:


rosella500:
There are a lot of commits in between and they're all me desperately trying to remove one commit that should be on a personal testing branch and not on Dalamud.

Well, except the last one. That's the one I mean to commit.

reiichi001 (approved):

PR #653, opened 2022-10-25
PartyIcons v1.0.9.3

squidmade:
Specific status icons now take priority over job icons.
- In a duty, the following icons are prioritized: Disconnecting, Viewing Cutscene, and Idle
- Outside of a duty, the following icons are prioritized: Disconnecting, Viewing Cutscene, Busy, Idle, Duty Finder, Party Leader, Party Member, and Role Playing

Thanks to Ces for a simple approach to this problem!
preview


squidmade:
I guess it can't build code that I never pushed

PR #658, opened 2022-10-26
[Testing] Beachcomber 1.0.2.1

rosella500:
- Crafts now display in the same language as the game. 
- Fixed bug where if you had exactly the number of materials a schedule required, it would still display as yellow


philpax (approved):
Fine by me, but I'd suggest renaming both the repo and internal name to reduce confusion in future

PR #660, opened 2022-10-26
[Testing] Better Shadows

Drahsid:
Initial submission for my BetterShadows plugin. This plugin allows the user to adjust the shadow cascade distances. Additionally, it has a small hack that the user can toggle (on by default) which will cause the 2048p shadowmap resolution setting to have a size of 4096p.


philpax:
bleatbot, approve

philpax:

philpax:
Looks like you're missing
`<PackageReference Include="DalamudPackager" Version="2.1.8" />`
in your csproj.

Koenari:
Note from a fellow dev: This PR is for testing/net6 and should probably go to testing/live. Afaik testing/net6 is retired.

MidoriKami:
You likely want to use `.net6-windows` instead of `.net6` 
https://github.com/Drahsid/BetterShadows/blob/master/BetterShadows.csproj#L4

This isn't the cause of an error, but it will make a lot of the build warnings go away.

MidoriKami:
You need to rebuild the project to update packages.lock.json

Drahsid:
Yeah, sorry I was using the dotnet command, but it seems that I needed to build with visual studio to update the package.lock

Drahsid:
Does `MakeZip` need to be true in `DalamudPackager.targets` ?

goaaats:
bleatbot, rebuild

reiichi001:
Hello! There are a couple pending changes I'd like to see before we merge this in. (The plugin has passed code review and rules review).

First, I'd recommend adding a config button for the plugin installer. This can be done by pulling in the UiBuilder service from Dalamud, if not already.
```
PluginInitializeMethod():
UiBuilder.OpenConfigUi += OpenConfigUiFunction;

Dispose():
UiBuilder.OpenConfigUi -= OpenConfigUiFunction;
```

Second, the configuration window needs some work, as parts of it were not reachable, and others are moving around strangely.
On initial install, it looked like this 
image
While that can be mostly fixed by expanding the UI window, I'd recommend adjusting it to a pre-configured size based on how large it should be to display all default options.

Next (and this is the actual blocker), the config window is clipping strangely.
image
All buttons aside from the preset names on the left move when the window is expanded, whether from dragging the column or by moving the window.
image
There are also buttons that are being clipped (Delete?) and (Unknown button shown almost entirely out of view). 
This should be fixed before the plugin goes live.

Drahsid:
I made the buttons try to align to the right of the column so that they would easily line up, that behavior was intentional.
I can't reproduce the clipping issue, but I think I just need to account for the padding in Imgui's settings to fix it.
The unknown button is just a "+" which adds your current settings as a new preset.
I'll push an update in a bit to try and resolve the issues that you pointed out.

MidoriKami:
This is just a suggestion but for copy save delete might be better to use icon buttons. You can imgui.pushid and imgui.popid to ensure the buttons have unique ids since icon buttons are a little weird.

Drahsid:
I did some reordering of the configuration UI. It shouldn't clip anymore.

philpax (approved):

philpax:
Apologies for the delay - thanks for submitting! You will be promptly added to the private channels on the Discord.

PR #664, opened 2022-10-28
DailyDuty v3.0.7.2

MidoriKami:
nofranz

Potentially fix a null reference exception on opening the duty finder.

This exception is non-fatal, and is handled gracefully currently. It also seems to only happen once per game session.


philpax (approved):
I think you might be able to use the null-coalescing operators here somehow

PR #691, opened 2022-11-02
[Testing] Beachcomber 1.0.3.3

rosella500:
- Allow for overwriting a day's value when it's automatically detected as a rest day (if advanced configuration option is checked)
- 	Note: This is a temporary feature while I work on getting more automatic detection in
- Fix slightly low workshop values in selection screen 
- Fix groove calculation when predicting future days

### DIFF
--- a/testing/live/Beachcomber/manifest.toml
+++ b/testing/live/Beachcomber/manifest.toml
@@ -1,8 +1,8 @@
 [plugin]
 repository = "https://github.com/rosella500/IslandWorkshopSolver.git"
-commit = "1f0be088043744d3e1dccb7f72bc4c68ab8cfb06"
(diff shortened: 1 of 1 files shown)


philpax (approved):
Is it possible for the `day` in `HasAllPeaksForward` to be greater than 7 and cause an infinite loop?

rosella500:
`day` can have values 1-7 (`CurrentDay` is % 7 and the value passed in is `CurrentDay + 1`), but if it were to be 8, wouldn't the loop `for(int i=0; i<7-day; i++)` just immediately terminate because 0 isn't less than -1? Unless there's a different loop I'm not seeing, I think it's fine. I'll add a check to early out to the next commit though.

PR #778, opened 2022-11-21
NoTankYou v5.0.3.5

MidoriKami:
Added a system that delays showing warning for 2 seconds after moving from one area to the next.

Eliminates random Summoner/Scholar warnings due to pets not loading in when the player is targetable.


MidoriKami:
Added fix to an out of range exception that occurs as PartyList is initializing

MidoriKami:
Fixed Logic Error, was reading just the seconds value of a timer, not the entire timespan causing flickering.

philpax (approved):

PR #788, opened 2022-11-25
[Testing] DeepDungeonTracker 0.0.0.1

marconsou:
New plugin Deep Dungeon Tracker.
Track things inside Deep Dungeons based on player action (Solo only).
Count how many kills, coffers opened, traps triggered, etc.
Basically, this plugin just reads data from network packets, system messages, and some from ObjectTable and presents the data for users as you see in the image below (there are other features):
Tracker
All collected data requires manual action from users.
This plugin can be useful for players which do some score runs or just want to check for some specific collected data.


philpax:

philpax (approved):
Thanks for submitting a plugin! If you're in the goat place Discord, let me know your Discord username and I'll have you added to the dev-private channels.

PR #793, opened 2022-11-26
XIVInstantMessenger 1.2.0.9

Limiana:
- Fixed a rare bug where a reply to incoming message could not be sent
- Fixed a bug where user could make window too transparent and it would disappear completely
- Replaced normal title flashing with gradient colors, making it better for eyes


philpax:
This will take some time to review. No guarantees as to when it'll be merged.

PR #808, opened 2022-12-01
[Testing] New Plugin: Mappy

MidoriKami:
Mappy is a imgui implementation of the ingame map with additional features.

This project is a work in progress, and many many more features are planned.

image
image


philpax:

MidoriKami:
Just completed a refactor of the original code, the original code did a lot of calculations manually using fields from lumina, however Caraxi pointed out there is a MapAgent that already contains the final calculated forms for a lot of these values.

The refactor lays out the system in two primary components, the MapManager, and the MapRenderer

Currently Implemented Features:
* Fates with bounding circles
Note: With fates that have a NPC that you speak with to start, these show the name and icon of the resulting fate, not the icon that the vanilla map shows. This is an intentional feature.

* Gathering Points
Note: Only gathering points located in the ObjectTable are shown, and they are not cached, so if the game removes a gathering point from the ObjectTable, it will no longer be shown by my plugin.
Note: I check the visibility of the gathering nodes to show only nodes that can be interacted with, this also has the side effect of triggering this feature when on the correct gathering job for a node

* Map Markers
Note: I do not render any text over the map, all information is provided via mouseover tooltips. I color the text of a tooltip for special markers.
Note: Clicking on an Aetheryte will use the Teleporter plugin to teleport to that Aetheryte.

I load the map texture and map icons in a different thread to prevent from causing framerate drops.
From my last set of testing, the Map Window itself uses approx 0.0300ms/frame for draw, and the plugin use 0.0020ms/frame for FrameworkUpdate. (Values determined using Debug build)
There may be peaks of up to 1ms when loading a new map from the Draw.

Given how much lumina data and math is being used, I believe the plugin to be performing very well =D

MidoriKami:
Now includes a richly featured configuration window/system.

This version has most of the planned initial features, minus ipc and quest markers.

https://user-images.githubusercontent.com/9083275/206434854-414562a4-453c-4a48-a0ba-dcb32427eb6b.mp4

MidoriKami:
These are the changes since the last manifest update:

https://github.com/MidoriKami/Mappy/compare/2003f12b0024313961e4a5f19441a710a0c4f14a..9e7c1fbd32f55186e1b33be461a507cefd65df21

philpax (approved):

PR #811, opened 2022-12-02
VFXEditor 1.7.1.5

0ceal0t:
- Added scd layout/track/attribute parsing based on Lumina
- Fixed issues with some scd files not being parsed properly
- Fixed issues when importing adpcm files
- No longer using obsolete hook creation method


philpax (approved):
you gotta make PRs more often if they end up being this big

PR #815, opened 2022-12-03
[Testing] Update WAIA - Major Refactor

reiichi001:
- fixes bug resulting in performance degradation.
- now allows a number to be at the start or end of a DTR bar entry.
- Franz actually uses the correct PR wording to make this post in the correct channel.


MidoriKami:
> Franz actually uses the correct PR wording to make this post in the correct channel.
Are you sure? Doesn't it need to be a capital T?

philpax (approved):

PR #828, opened 2022-12-09
[Testing] Mappy v0.3.0.0

MidoriKami:
`Configuration Window` - Add information about why Mappy cannot render below the native game UI
`FATEs` - Add Option to color FATEs that are about to expire with a configurable early warning time and color
`MapMarkers` - Improve Performance
`Map Window` - Add "Hide Between Areas" option to hide the map during zone changes, and automatically reshow it afterwards
`Map Toolbar` - Fix Scaling Issues

`Integrations` - Added option to redirect vanilla map functions to Mappy
These integrations ***replace*** the vanilla behavior.
If a flag would be placed on the vanilla map, Mappy opens and places a flag.
If a gathering marker would be placed on the vanilla map, Mappy opens and places a gathering marker.
If a hotkey or button or command would open the vanilla map, Mappy will open the plugin map.

These integrations will automatically and forcibly disable upon entering PvP.
They will re-enable according to your settings upon Leaving PvP.

Known Issues:
Missing Housing Icons
Missing Quest Icons (both claimed quests, and unclaimed quests)

### DIFF
--- a/testing/live/Mappy/manifest.toml
+++ b/testing/live/Mappy/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/Mappy.git"
-commit = "9cdf65c2cb9013a13764bea74a361e1255d5ff79"
+commit = "dfa4aca3a3a62a70c8c7c54f189402c663cb13cd"
 owners = ["MidoriKami"]
 project_path = "Mappy"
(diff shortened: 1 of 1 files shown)


MidoriKami:
Misc Changes: 
Fixed calculation error that this version introduced.
Removed Auto-open from DebugWindow and ConfigurationWindow

MidoriKami:
bleatbot, rebuild

MidoriKami:
Forgot to hit push =x

MidoriKami:
Copy-paste error, put in wrong commit ID

MidoriKami:
That's the final edit for this PR.
I just changed the default value for the integration features to on.

All the hooked functions are void return, so its extremely unlikely there will be any stack corruption, additionally everything hooked has exception handling configured. There should be no realistic user risk to having these integrations on by default.

If a user doesn't want the integrations on, they can disable all of them at once with a single prominent checkbox in the configuration window.

karashiiro (approved):

PR #837, opened 2022-12-11
[Testing] WAIA 0.0.1.3

reiichi001:
Fixes an issue wherein adjusting the config string would not update the DTR bar.

DTR length has also been limited to 35 characters. In the event that you load every template, you may hit the limit.


MidoriKami:
To clarify only the formatting string itself is limited to 35 characters. The actual dtr entry length is not limited. Theoretical max length is 4x the longest PlaceName in the game + 23 characters

philpax (approved):

PR #843, opened 2022-12-12
DeepDungeonDex 2.0.2

wolfcomp:
### Bug Fixes

* null reference with legacy window when mob note is not set (4eab828)


wolfcomp:
Gona push 2.0.3 out instead

PR #860, opened 2022-12-18
GatherBuddy 3.2.0.0

Ottermandias:
- Add Ellipses for fishing spot names in the fishing timer, if the name is too long (by quisquous)
- Prevent some gather actions from occuring during loading screens or when not logged in.
- Add hover tooltips to fishing timer (they are only available when clickthrough is not enabled)
- Add some support for ocean fishing. Rudimentary check for ocean times for the fish tab and availability-checks for current weather/time per route. This was mostly done by quisquous and was not tested by myself except for some rudimentary checks and looking through the code, since I don't ocean fish. Please let us know if something doesn't work correctly.
- Some fish data updates.


philpax (approved):
that is fascinating formatting in the Data file

PR #864, opened 2022-12-19
 Player Tags v1.8.4

Pilzinsel64:
- A few High-DPI fixes for config UI
  *Please let me know, if I missed some UI elements. Normally I don't use anything else then 100%.*


MidoriKami:
Dalamud provides helpers `ImGuiHelpers.GlobalScale` and `ImguiHelpers.ScaledVector2( ... )` for doing scaling stuff easier.
Not saying you need to use these, but if you come across weird issues it's worth considering them.

Pilzinsel64:
Uh, thank you a lot @MidoriKami for the hint! That are things I simply can't know but really are looking for instead doing my own crappy code.

PR #871, opened 2022-12-22
[Testing] Mappy v0.4.6.0

MidoriKami:
`Flag Marker` - Fix Adjustments using wrong label for scale
`Gathering Area Marker` - Fix Adjustments using wrong label for scale
`Focus Layer` - Add internal module for displaying on top of all other modules
`Focus Layer` - Add Aetherytes display to focus layer when `Draw Aetherytes on Top` is enabled
`MapManger` - Adjusted viewport logic when changing maps
`MapManger` - Now saves and restores viewports when returning to a map you have looked at before
`Penumbra` - Use InterfacePath instead of DefaultPath for texture loading

I am taking a break from working on Mappy. I have spent well over 40 hours in the last week attempting to get the necessary data from the game to make quest markers work correctly, while at the same time constantly being bombarded with "xyz doesn't work" for the very things I have been working on or "make mappy work like vanilla" comments when Mappy is was never intended to copy the vanilla map, but to be an improved version. 

I no longer have the drive to continue working on this project at this time.

**No support for Mappy issues will be provided**

### DIFF
--- a/testing/live/Mappy/manifest.toml
+++ b/testing/live/Mappy/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/Mappy.git"
-commit = "6b42ba757c0768b992f7a636c42e007e1e859226"
+commit = "7aaf43a06da5b100d2859b5358e01a60f51b3adb"
 owners = ["MidoriKami"]
 project_path = "Mappy"


MidoriKami:
bleatbot, rebuilt

MidoriKami:
bleatbot, rebuild

reiichi001 (approved):

PR #905, opened 2023-01-03
[Testing] DailyDuty v3.2.0.1

MidoriKami:
Localization Update

There will likely be errors in translation, please submit corrections to <https://crowdin.com/project/dailyduty>

KamiLib needs translating, there are not a lot of strings, so it should be pretty easy: <https://crowdin.com/project/kamilib>


MidoriKami:
This may seem like a huge PR, but it's entirely localization changes, from using a custom static string class based system to using resx. This required every string reference to be renamed to the new format.

reiichi001 (approved):

PR #924, opened 2023-01-08
[NET7] ChillFrames v1.6.1.0

MidoriKami:
New commands have been added use `/pcf help` to view all commands
This update will reset your zone blacklist


MidoriKami:
No code changes since last testing release

PR #927, opened 2023-01-08
[NET7] DailyDuty v3.2.1.0

MidoriKami:
New commands have been added use `/dd help` to view all commands


MidoriKami:
Removed custom login handler, using Dalamud OnLogin to check for localplayer

PR #937, opened 2023-01-09
[NET7] Distance 1.0.1.1

PunishedPineapple:
- Added options for distance-based coloring on nameplates.  This probably isn't as useful as you think it is, but it's there for the people that wanted it.
- Increased maximum screen position offset for individual widgets.
- Updated for Dalamud API 8.


reiichi001:
bleatbot, rebuild

PR #938, opened 2023-01-09
Release: BetterMountRoulette 1.0.1.6

CMDRNuffin:
Features:
- Add mount groups
- Associate each mount roulette with a separate mount group (or none at all)
- Summon a mount from a specified group via /pmount <group name>


CMDRNuffin:
Never mind, I'll just release it as 1.0.1.7 for .net 7

PR #1002, opened 2023-01-11
[Testing] RadarPlugin

KangasZ:
This is an initial PR to add this plugin to the public repository. This is partially to see whether the community is ok with this type of plugin, even though a functionality nearly identical to this plugin is already included in dalamud dev menu.

I don't personally think the plugin provides a combat or other advantage over other players at it's base functionality, but does increase visual acuity to an extreme, especially in ffxiv content where mobs are scary (deep dungeons and eureka).

### DIFF
--- a/stable/AdaptiveHud/manifest.toml
+++ b/stable/AdaptiveHud/manifest.toml
@@ -1,8 +1,8 @@
 [plugin]
 repository = "https://github.com/Helios747/AdaptiveHudPlugin.git"
-commit = "b078d01f2ca05bb069e897659c79136a48d341a8"
+commit = "3245104ecad6b2e9db2cd3963c7458231f26a666"
 owners = [
     "Helios747"
 ]
 project_path = "AdaptiveHud"
-changelog = "Fix config not being read"
+changelog = "API 8"
--- a/stable/ChatTwo/manifest.toml
+++ b/stable/ChatTwo/manifest.toml
@@ -1,10 +1,7 @@
 [plugin]
 repository = 'https://git.anna.lgbt/ascclemens/ChatTwo.git'
-commit = '38b599b684bc73b1ec932b084ca1ed8e82b25ddc'
+commit = '53043bd2864ff29b26e4440b90106f9fc5a6abb9'
 owners = [ 'ascclemens' ]
 changelog = """\
-- Updated translations.
-- Changed some internals for (minor) improved performance.
-- Fixed an issue with empty lines causing some flickering.
-- Fixed an issue with empty lines disappearing under some circumstances.
+- API 8
 """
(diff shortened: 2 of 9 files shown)


MidoriKami:
Please include a description here of what the plugin does, maybe also attach pictures if you have any

KangasZ:
This plugin is intended to be an easy way to view and see players, mobs, and other entities in the world via a 3D-radar interface. It is intended to be relatively customizable and intends to block out entities that would not normally be able to be shown, like invisible mobs. It was created originally to easily see mobs in POTD/HOH and Eureka. I have started added checks such that it will not show in PVP areas, so as to mitigate the use of it in PVP - although I don't believe, if used, would even provide a significant advantage like an FPS game.

There is no polling of the game servers in any way and everything used is from the Dalamud's GameObject class, beside one manual check on whether or not it has any render flags from FFXIVClientStructs GameObject. All the data is obtained via looping through the ObjectTable then displaying it in ImGui background list. If you have any specific questions about implementation or usage, feel free to check source or reply/let me know here.


image
image

reiichi001:
bleatbot, approve

MidoriKami:
Regarding invisible mobs, `FFXIVClientStructs.FFXIV.Client.Game.Object.GameObject` has a method `GetIsTargetable()` that can filter out invisible objects.

wolfcomp:
For all PvP areas you have `ClientState.IsPvP` this includes wolves den.

MidoriKami:
Additionally, I think it would probably be fine to update the overlays every frame, we have a huge amount of performance space we can use for plugins. Looping through the object table seems like it would be expensive at first, but it really isn't.

KangasZ:
`GetIsTargetable()` seems to filter out some other stuff like aetherytes. May also filter out other things I want displayed ie Auspices in HoH. Would have to investigate deeper or make it a config option.

Updating the table once per second was done earlier in development but I don't particularly see a need to update it to each frame. Updating mob info takes .1-.3ms at most loads, where the typical draw time is .05-0.15ms. I could see reducing the amount of times per second it refreshes, but don't like adding additional load bound to frames.

Added the IsPvP check. Will update the pr with the correct commit in a bit.

philpax:

philpax:
Hi there,

Sorry about the wait. We discussed this significantly, especially given the existence of similar plugins. Unfortunately, we've come to the conclusion that we have to reject Radar.

The primary rule from [our guidelines](https://goatcorp.github.io/faq/development#q-what-am-i-allowed-to-do-in-my-plugin) that Radar contravenes is:

> your plugin does not augment, alter, or interfere with combat, unless it only provides information about your own party or alliance members that is otherwise available, but represents said information differently. 

Fundamentally, it provides information about combat participants that are not your party and alliance members. There's a fair bit of content that's reliant on mobs sneaking up on you / not being visible; Radar would directly impact the difficulty of that content by showing them to you when you would not normally be able to see them. 

Our apologies!

PR #1030, opened 2023-01-12
[Testing] Hybrid Camera

Drahsid:
This is a plugin that is similar to the strafing tweaks in simple tweaks, but is a bit more pedantic. It solves a couple more issues by letting the user switch their movement mode with particular keypresses. The end result is a bit more freedom of movement, while retaining the ability to rotate the player character when idle, like in standard movement.
Additionally, this was preemptively updated to API8/NET7.


Drahsid:
Fixed a bug with out how I was initializing the config

philpax:

Drahsid:
The versioning issue mentioned in Pull Request #1191 also existed here. I fixed it in this last update.

philpax (approved):
Sorry about the delay on this one; we had concerns about it resulting in a noticeable change in movement. We've since tested it and it seems to work OK, but we'd suggest smoothing the transition if possible.

PR #1051, opened 2023-01-13
Player Tags v1.9

Pilzinsel64:
* Target to API 8
* Target to .NET 7
* Updated translations


Pilzinsel64:
Urg I always forget to change the commit, it's mostly a "nah, do the changelog first" and then forget. :D

PR #1076, opened 2023-01-14
[Testing] MarketUploader

edg-l:
This pr adds the initial release for the MarketUploader plugin on testing.

Uploads market data to market.xivhub.org and other configurable aggregators.


reiichi001:
New plugins may not be PRed directly to Stable. Please start with only testing. I've also added the `[Testing]` prefix to this PR, so it will show in the correct section.

After that, this plugin will be added to the review queue.

edg-l:
> New plugins may not be PRed directly to Stable. Please start with only testing. I've also added the `[Testing]` prefix to this PR, so it will show in the correct section.
> 
> After that, this plugin will be added to the review queue.

done

philpax:

philpax:
Hi, sorry about the delay! This actually sparked a little bit of a discussion amongst ourselves - we realised that we shouldn't be uploading local content ID in Dalamud, and we shouldn't let plugins do it either. Is it possible for you to switch to `hash(localcontentid)` or similar so that there's no way to recover the original ID? Once that's done, we're happy to merge :)

edg-l:
done

philpax (approved):

PR #1081, opened 2023-01-14
[Testing] Updates MonsterLootHunter 1.0.3 to 1.0.4

danielbrenom:
Adds search term on command
Updates API and NET version bump


reiichi001:
Stable will override Testing if it's newer. This PR isn't needed. Testing releases should only include newer-than-stable releases explicitly for testing. Once a place it stable, the folder from testing can be removed.

PR #1115, opened 2023-01-16
DailyDuty v3.2.1.5

MidoriKami:
Removes button that people apparently struggled to find in the first place.
It likely won't be missed.


MidoriKami:
This changes how Normal Raids and Alliance Raids reset themselves. It used to be triggered by a button, now it's automatic.
The automatic behavior comes with default settings, so if the user doesn't notice that's on them.
I got tired of people not seeing the completely obvious "reload duty info" button and requesting that I add the new alliance raid.

reiichi001 (approved):

PR #1120, opened 2023-01-16
[Testing] XIVWindowResizer v1.0.0.0

AlexCSDev:
This plugin adds command which allows to resize FFXIV window to arbitrary size/resolution. This is primarily intentended to be used to take screenshots in higher resolutions than display resolution without using DSR and other supersampling technologies because those usually require you to switch to fullscreen mode and that might be very annoying when you are using external applications like Anamnesis.

Since this is quite simple plugin and similar functionality (just implemented a bit differently) has been used in XIVR for quite a while now, I don't think there is any need to push it to testing first.


reiichi001:
All new plugins should be pushed to testing first, even if they're using slices of other plugins. After a testing period of a few days, it can then be moved to stable.

reiichi001:
bleatbot, approve

AlexCSDev:
Alright, I assumed that this is not strict requirement based on what I've read in docs, moved into testing.

philpax:

philpax (approved):

philpax:
 
AlexCSDev:
Thank you, my discord handle is: Alexsey#9211

PR #1141, opened 2023-01-18
[Testing] MinionRoulette

InitialDet:
The idea of this plugin is very simple, every time the player change zones, their current minion is swapped with a random one.


InitialDet:
I've made this plugin quite a while back but kept it to myself because of the way its designed. Personally i think this is quite an innocent and fun plugin , but i believe it falls under the botting category since it uses Minion Roulette (action) when a zone change is detected and the player minion is currently summoned.

That being said, I decided to make this PR to get an official response if this acceptable or not, just in case.

philpax:

philpax:
Hi there,

Sorry about the wait; this really did take a fair bit of back and forth on our part, as well as discussion with the community. Unfortunately, as you might have already guessed, this is automation - no matter how harmless it is - and any attempts to make it less detectable would be behaviourally dubious. As a result, we have to reject it.

We would be happy with a variant that shows a button that the user can click to change their minion when the zone changes (or to highlight the existing button on the hotbar), but anything that would automatically act is [outside of our guidelines](https://goatcorp.github.io/faq/development#q-what-am-i-allowed-to-do-in-my-plugin).

Our apologies!

PR #1164, opened 2023-01-20
[Testing] Wholist v0.1.0.0

BitsOfAByte:
- Performance improvements
- Removed 'Filter Suspected Bots' and integrated that behaviour by default


BitsOfAByte:
Note to reviewers - a lot of plugin behaviour has been moved into a commons library as it was/will be shared between a lot of my plugins.

philpax (approved):

PR #1167, opened 2023-01-20
DamageInfoPlugin 2.2.0.0 -> 2.2.0.3

lmcintyre:
- Fixes flytext damage type mismatches by using network data, the same method that SE uses to display their damage type icons
- Inaccuracies should be report with an accompanying dalamud.log file in the Damage Info thread in Discord


lmcintyre:
bleatbot, rebuild

reiichi001 (approved):

PR #1170, opened 2023-01-21
DailyDuty, NoTankYou, ChillFrames Fixes

MidoriKami:
nofranz

There is an issue with file access sometimes for my plugins that use custom file saving and loading.

This fixes the `The process cannot access the file` error that sometimes occurs.


MidoriKami:
Closing to investigate potentially new issues caused by these changes.

PR #1181, opened 2023-01-22
[Testing] OofPlugin 1.1.0.0

codelastnight:
- Add Hbomberguy video
- Oof on Party Member's death
- Oof on Alliance member's death (maybe works, needs testing and a refactor)
- load custom audio file
- update to API 8, .net7


goaaats:
I'd recommend using Util.OpenLink() instead of Process to show browser windows, looks good otherwise though.

goaaats (approved):

PR #1183, opened 2023-01-22
[Testing] TargetLines plugin

Drahsid:
A plugin that I threw together in a couple hours which visualizes target lines in ImGui. I added a bunch of customization options. It is kinda like peeping tom, but for all players, and enemies. Probably doesn't provide any advantage for actual gameplay.

### DIFF
--- a/testing/live/TargetLines/manifest.toml
+++ b/testing/live/TargetLines/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/Drahsid/TargetLines.git"
-commit = "1ad7e11ccb894c079ad0bbe38d1c15871ca6052f"
+commit = "932b1fa8c42ff849aa58deba05c53d2f8a497a4a"
 owners = ["Drahsid"]
 project_path = ""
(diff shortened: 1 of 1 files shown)


Drahsid:
Some worries were brought in in the Discord over the possibility that this could show target lines from Objects that are not visible and not tangible, thus providing an advantage in comnat. I have updated the plugin to prevent that by doing the following:
- If the Object is not targetable, it will not draw target lines from it
- If the Object is a BattleChara, and not a PC, and it is occluded from the camera, it will not draw target lines from it. I do this check with a raycast and by checked the the tangent between the camera's direction and the Object's position relative to the camera position (if it is behind the camera.)
- If the Object has a scale of 0

This seems to work in my quick tests. Let me know if there are any issues (worst case scenario I disable it entirely during combat.)

Drahsid:
The versioning issue mentioned in Pull Request #1191 also existed here. I fixed it in this last update.

philpax:

philpax:
Hi, sorry about the delay. We're happy to merge, but can you make sure it's disabled in PvP?

Drahsid:
Oh shoot, I didn't even test it in PvP. I'll update it shortly and let you know when that happens.

Drahsid:
I just pushed the update. Let me know if there are any other issues.

philpax (approved):

PR #1186, opened 2023-01-22
JobBars v1.2.3.4

0ceal0t:
- Fixed issues with icons being improperly replaced
- Fixed duration of living shadow
- Fixed divine veil cooldown tracker

### DIFF
--- a/stable/JobBars/manifest.toml
+++ b/stable/JobBars/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/0ceal0t/JobBars.git"
-commit = "c269978471b9910c04e2501229790df163a7cea1"
+commit = "ec2de380083edecc0b6f5bf9479d5c943511dcae"
 owners = [
     "0ceal0t",
 ]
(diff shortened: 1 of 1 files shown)


goaaats:
Did you always have binaries for client structs in here? Could you use a submodule instead? If not, why?

MidoriKami:
When 6.3 released, I cloned built and tested Jobbars locally, and removed the reference to ClientStructs and used Dalamuds without issue.

0ceal0t:
> Did you always have binaries for client structs in here? Could you use a submodule instead? If not, why?

yeah, there were always binaries in there, since sometimes I've need to make custom changes before they've been merged into the main repository

philpax:
blocked on @goaaats's call on what to do

0ceal0t:
I mean I can remove them since right now CS has everything I need

philpax:
oh sweet then yeah that would be great

philpax:
pingerino

0ceal0t:
@philpax I removed the CS binaries

philpax (approved):

PR #1188, opened 2023-01-22
[Testing] HudManager v2.5.10-b1

zacharied:
* Fix job gauges thinking that the game is in gamepad mode even when in mouse mode

There is currently a known issue where entering the Character Config menu and changing from Mouse Mode to Gamepad Mode, and then back to Mouse Mode, will cause swaps to stop working until the HudManager menu is opened again. Try to avoid doing this.


goaaats (approved):

philpax:
In future, I suggest using the GitHub web editor to edit the manifests; that'll ensure you're always editing the latest manifest, and you won't have to worry about merging / syncing your fork of this repo.

PR #1191, opened 2023-01-23
Better shadows Update 4

Drahsid:
Added per zone config options.
Added /tbshadows for toggling the hacks on the command line
Hacks themselves are unchanged.


goaaats:
You need to make sure that your plugin versions are sequentially increasing - otherwise, people will not get the update. This happened before with another one of your plugins.

Drahsid:
You're right. It looks like I misunderstood how the assembly version automation that I was using worked. It should now consistently produce a higher version number.

philpax (approved):

PR #1192, opened 2023-01-23
Add GentleTouch 1.6.3

fitzchivalrik:
- fix: Do not vibrate on already attuned Aether Currents


philpax (approved):
I wonder if `ObjectTable` could be LINQ-able

fitzchivalrik:
It's too long, but I remember trying to optimise the loop, not using any LINQ constructs or similar was faster, I think, mainly because of fewer allocations.

PR #1193, opened 2023-01-23
Add Compass 1.26.3

fitzchivalrik:
- chore: Update to API 8

Dalamud's updated too; the crashing feature should be removed.


philpax (approved):
your formatting style makes me wonder if you've been writing Haskell 🤨

fitzchivalrik:
I use Rider at work, and the style used there was apparently (partially) synced via the settings sync. Easier to just live with it than figure it out...

PR #1197, opened 2023-01-24
ChillFrames v1.6.1.3

MidoriKami:
Fix framerate target being applied while you are typing
Add disable during bard performance


MidoriKami:
wait, don't merge... one minute.

MidoriKami:
There... forgot to actually add it to the ui to enable the feature tehe

karashiiro (approved):

PR #1209, opened 2023-01-27
[Testing] Hunty

Infiziert90:
Is it finally time to finish your hunting log???

### Hunty

+ List of monsters in the same location based on Rank
+ Clickable coords to open your map on that exact location
+ Easy to switch between Grand Company and Jobs
+ Automatic selection based on current job

### Image

hunty


reiichi001:
The class/GC button is getting clipped regardless of window width. 

image

Infiziert90:
> The class/GC button is getting clipped regardless of window width.
> 
> image

Should be fixed in the newest update, calculating the text size now instead of fixed value ... forgot to change that before PRing

philpax (approved):

philpax:
Sorry about the delay! In the interests of getting this through the door, I've merged it, but one of our reviewers mentioned that it would be nice if you could use Dalamud's newtonsoft json instead of the NuGet version (just to avoid having additional assemblies floating around))

Infiziert90:
> Sorry about the delay! In the interests of getting this through the door, I've merged it, but one of our reviewers mentioned that it would be nice if you could use Dalamud's newtonsoft json instead of the NuGet version (just to avoid having additional assemblies floating around))

Newest version is using dalamud's newtonsoft json

PR #1212, opened 2023-01-27
[Testing] DeepDungeonTracker 0.0.2.0

marconsou:
- [**Breaking Changes**] All save files up to this version need to be converted (go to the Statistics Tab and scroll down to [Convert Save Utility] section for more details).
- Added the [Convert Save Utility] section on the Statistics Tab.
- Added a date and time stamp to the file name when creating backup files (avoid overwriting files).
- Added a delete confirmation dialog before deleting a backup file (avoid misclick).
- Pomanders (Got and Used) in the Total section (Statistics Window) will now be automatically sorted (better visualization).
- Fixed a rare issue in which a save file was not properly deleted and then starting a new save file on the same slot caused both data files to be merged.


MidoriKami:
If someone doesn't convert beforehand will they crash?

marconsou:
No, basically some data will be shown using deprecated icons.

MidoriKami:
Ah alright. It's rather annoying to do but you can manually convert configs using JObject, that's what I do in NoTankYou and ChillFrames, 

https://github.com/MidoriKami/ChillFrames/blob/master/ChillFrames/Config/ConfigMigration.cs
https://github.com/MidoriKami/KamiLib/blob/master/Configuration/Migrate.cs

marconsou:
I don't think it will be needed because the data that needs to be updated is not from the configuration file (where it will be loaded at start and is just one file).

The data to be updated is present in one or more files, and it will only affect files stored right now on the user's computer. 
For newly created files, it will not be an issue.

Not knowing how many files the user has (can be 1 or several files), trying to do this process automatically by scanning for those files would (potentially) cause more issues. I don't have a flag to indicate if a file needs to be updated; even if I did, it would result in some extra (and unnecessary) reading of those files, which I don't think is a good solution for this case.

So I opted (for this case) to do it manually (the user just needs to press one button in the file they want to fix) as a temporary measure.

MidoriKami:
Makes sense, and seems entirely reasonable.

philpax (approved):

PR #1217, opened 2023-01-28
DeathRoll 1.0.0.3

Infiziert90:
+ New timer argument which toggles the timer (/drh timer)
+ New option to reset all rolls on timer start (default false)

If you encounter any issues with roll parsing (e.g not picking up other players and instead only you), please activate the debug option, try to reproduce and send your dalamud log~


goaaats:
bleatbot, rebuild

philpax (approved):

PR #1219, opened 2023-01-28
[Testing] Better Targeting System

Lynesth:
This plugin tentatively aims at improving the tab targeting in the game by following simple rules and using fallbacks (like targeting an enemy in your enemy list if there's nothing to target on screen, etc).

It doesn't - or at least shouldn't - allow you to targeting anything not targetable the vanilla way.

It also adds a keybind to target the lowest health enemy visible on screen.

`/bts` provides options to configure the keybinds.

### DIFF
--- a/testing/live/BetterTargetingSystem/manifest.toml
+++ b/testing/live/BetterTargetingSystem/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Lynesth/BetterTargetingSystem.git"
-commit = "4459d1db1bfa8d445cc1e52be95b47695abc9a8f"
+commit = "27ffb252fd187200c9220c188611269e773b846d"
 owners = ["Lynesth"]
 project_path = "BetterTargetingSystem"
 changelog = "Initial submission"
--- a/testing/live/BetterTargetingSystem/manifest.toml
+++ b/testing/live/BetterTargetingSystem/manifest.toml
@@ -1,0 +1,6 @@
+[plugin]
+repository = "https://github.com/Lynesth/BetterTargetingSystem.git"
+commit = "fc570e8affc5dd7f5b570ea1f1b87a949caa25c7"
+owners = ["Lynesth"]
+project_path = "BetterTargetingSystem"
+changelog = "Initial submission"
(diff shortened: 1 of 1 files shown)


goaaats:
bleatbot, approve

Lynesth:
Okay, unless I'm asked by you or if I find a bug in the meantime, I won't be touching this anymore until it's been reviewed and (hopefully) approved or rejected.

Feel free to contact me on the discord server as well if you want to chat directly about it.

philpax:
Hi, sorry about the delay - it took a bit of time for us to make a decision because we had to consider what extent this "thinks for you."

I'm happy to say that it's fine and we're approving it, but can you make sure that it's fully disabled in PvP? Otherwise, it's good to go! :)

goaaats:
bleatbot, rebuild

Lynesth:
@philpax No worries about the delay, and thanks for your answer.

Regarding PvP, am I required to disable the plugin entirely or would just disabling the extra non-vanilla keybinds would be fine?
I personally think the changes on [Cycle Targets] don't give an advantage as it doesn't use PvP specific addons to target enemies and such. And getting used to the way your targeting works in PvE to then have to deal with another behavior in PvP doesn't make for a nice experience either.

But I will nonetheless abide by your decision, just let me know. :)

philpax:
Sorry once again for the late reply. We'd prefer if you disabled it entirely in PvP for now - we're concerned that its lowest-health functionality might result in an advantage in player combat. We can re-evaluate later if that ends up not being an issue :)

Lynesth:
@philpax If this is your main concern, then know that the `Target Lowest Health` and the `Target "Best" enemy for AoE` are the non-vanilla keybinds I am referring to in my previous comment and are already disabled for PvP in the repo.

Only things not disabled are `Cycle Targets` and `Target Closest` which both are simply tweaked from their vanilla counterpart but without being intelligent in what it picks for you, just trying to be more consistent with what the player would expect when pressing said keybind (and since there's no enemy list in PvP, they are even less useful).

mspaint_2023-02-16_20-41-58

philpax (approved):

PR #1221, opened 2023-01-29
[Testing] Add GentleTouch 1.10.0

fitzchivalrik:
- feat: DualSense support via DS4 compatibility vibrations
- feat(DualSense): Set resistance for Adaptive Triggers
- feat(DS+DS4): Two extra macro buttons: Create (DualSense) / TouchPad (DualShock4) as Individual Macro Nr 96, PS Button as Individual Macro Nr 97
- feat: Option to /draw & /sheathe with PS button instead of Macro Nr 97

Check out the new settings tab, if you are using a DualSense/DualShock4.


philpax (approved):
TODO(philpax): review this PR

PR #1230, opened 2023-01-29
[Testing] TF2-ish Critical Hits 1.1.0.0

Berna-L:
Now, you can preview your sound file directly in the configuration pane.

Also, fixes not playing the sound at all (which is, like, 75% why someone would install this plugin).


Berna-L:
Actually, I think I'm going to push some QoL features. Hold on...

philpax (approved):

PR #1251, opened 2023-01-31
[Testing] Allagan Tools v1.3.0.2

Critical-Impact:
Stop uncraftable items from being added to craft lists(also allow them to be deleted if you've run into the issue) 
Replace memory sort scanning with ODR parsing(should still be relatively fast and keep things up to date) 
Fixed some other issues
Replaced some imgui code with what I'm told is the more correct way of doing it, also added a EndChild that was missing.


Critical-Impact:
To whoever reviews this, the changes in most of the "Window" files are just me removing the if statements I have on BeginChild calls. Apparently this isn't the right way to do things(as far as I can see of the demo code and from what people say on discord)
Obviously still have a look through but I tried to change nothing else bar the removal of the if statement then having Rider realign the rest of the code.

PR #1256, opened 2023-02-01
[Testing] DailyDuty v3.2.2.3

MidoriKami:
`Raids (Normal)` - Remove The Omega Protocol (Ultimate) for showing
`Hunt Marks (Weekly)` - Remove "Force Complete" option, use suppression instead
`Configuration` - Fix plugin failing to load on first install


MidoriKami:
bleatbot, rebuild

PR #1259, opened 2023-02-01
DelvUI release 1.5.3.0

Tischel:
- Reworked name text tags:
    + A lot of existing text tags have been removed.
    + You can now specify the type of unit for a name text tag.
        * This allows to display player names and npc names differently on the same label.
        * Example: `[player_name:initials][npc_name]` would display the initials for a player, and the full name for an npc.
        * You can still use tags starting with `[name:` that work for both types of units.
    + Additionally you can now specify a cap for the length of the names by adding `.#` at the end.
        * Example: `[name.5]` would display `Tisch` for the name `Tischel`.

- Disabled Sign Icons from unit frames for non-combat NPCs (until I find a better fix).

### DIFF
--- a/stable/DelvUI/manifest.toml
+++ b/stable/DelvUI/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/DelvUI/DelvUI.git"
-commit = "b2cd93f6d5d67a5bf3db684081ddca0f3386d7e6"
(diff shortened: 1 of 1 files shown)


Tischel:
Bad bot

NotNite:
What are you trying to do closing and re-opening the PR? I'm confused.

philpax (approved):

PR #1267, opened 2023-02-04
Update EngageTimer to v2.2.5.1

xorus:
- Add setting for display threshold: allows you to specify when the timer will become visible (can be useful if you don't want to see numbers irrelevant to your pre-pull)
- Awkded new number styles


philpax (approved):
lol

PR #1280, opened 2023-02-06
[Testing] ClarityInChaos v1.1.2.0

meoiswa:
Clarity In Chaos is a simple plugin that allows the user to configure and automatically apply different Battle Effects settings according to Party size.
image


goaaats:
bleatbot, rebuild

philpax:
@Lerofni tested ClarityInChaos and found it to work well, but noticed two things:

- Changing BattleEffects via the native UI doesnt seem to work and it always jumps back to what is currently set in the plugin
- The Debug Messages enabled by the Debug Messages option print to chat rather than xllog

Would you be able to fix these? With the former one, we'd prefer you respect the native setting, but if you're unable to do so we'll take an in-UI warning that it overrides the native control. Thanks!

meoiswa:
- Respecting the in-game settings would complicate the solution, it can be done and I'll work on it for a future release, but I'll place an in-UI warning for now.
 - The intention of the Debug Messages on chat is for users of the plugin to better understand what their settings do and when they change. I'll change the logic so that it always logs to the Logs (for developers), and only displays chat messages when the setting is enabled (for users), along with changing the UI so that it better reflects this behavior.
image

PR #1282, opened 2023-02-06
MemoryMarker v0.1.0.2

MidoriKami:
New Plugin "MemoryMarker"

A simple Waymark Management plugin with no configuration.
Automatically saves and loads waymarks when entering and leaving zones, so you get 30 waymarks **per-zone**.
Allows you to edit how the waymarks display in the Waymarks menu.

If WaymarkPreset is also installed, the auto-load feature will automatically be disabled.

image


MidoriKami:
This is a release from testing to stable with no changes made from the testing release.

lmcintyre (approved):

PR #1285, opened 2023-02-06
[Testing] Fungah

imaginary-png:
Just draws a lil dot on the south safe spot during Any Way the Wind Blows, to make it easier to see during big crowds.
image1
image2


imaginary-png:
bleatbot, rebuild

imaginary-png:
:(

imaginary-png:
bleatbot, rebuild

imaginary-png:
it is? isn't it?

Infiziert90:
Your internal plugin name is still "Sample Plugin" and I believe Dp17 uses that one for paths

imaginary-png:
bleatbot, rebuild

goaaats:
bleatbot, rebuild

Infiziert90:
Tested your plugin for the PAC as v0.0.0.1 and noticed some smaller issues/slight annoyances for myself at least.

+ The point isn't static, it moves quite a bit
+ The point is always there, you just have to be close enough to the arena for it to be triggered

Both of them combined makes it a bit distracting and maybe also misleading as other games are played in the same arena.
So making it only appear while duty bound with the current duty / entered duty being the specific game would help a lot.

PR #1300, opened 2023-02-09
[Testing] Mounting Party v1.0.0.0

meoiswa:
This simple plugin overrides Mount Roulette to spawn a randomly picked multi-seater mount the user has unlocked, whenever they are in a Party

### DIFF
--- a/testing/live/MountingParty/manifest.toml
+++ b/testing/live/MountingParty/manifest.toml
@@ -1,0 +1,8 @@
+[plugin]
+repository = "https://github.com/meoiswa/MountingPartyPlugin.git"
+commit = "572d44d8732f0166e3639171d21dcb7f7448eebc"
+owners = [
+    "meoiswa"
+]
+project_path = "MountingParty"
+changelog = "Initial Release"


goaaats:
bleatbot, rebuild

philpax:
@tommadness has tested this and has noted that it conflicts with Better Mount Roulette. Is there a reason why you've created a new plugin instead of contributing to BMR? If so, is that something we can help with? The two plugins overlap significantly, so we'd like to reduce the amount of duplication if possible.

meoiswa:
No reason in particular, I just didn't feel it was quite the same functionality, as BMR is a lot more customizable.
If you guys think it should be a feature of BMR i can just make it a PR

CMDRNuffin:
I am currently in the middle of a major refactoring (including migrating to per-character config), so it would probably make sense to do that after (in a way that maybe doesn't _totally_ obsolete this plugin, maybe via ipc?), and in the meantime just have a warning that these plugins don't currently work together. Or put this on hold altogether.

philpax:
Our preference would be as part of BMR and/or to make this functionality easily accessible through BMR, so that people who want the simple experience can get there quickly. If you would prefer to keep it as a separate plugin, that's fine, but it does mean that you'll have to deal with BMR interop problems.

meoiswa:
As suggested, I've turned this plugin into a feature of BMR, currently PR'd [here](https://github.com/CMDRNuffin/BetterMountRoulette/pull/9)

PR #1301, opened 2023-02-09
[Testing] Neko Fans 1.2.3

Meisterlala:
Not an update, nothing about the code changed. Stable and Testing are now the same.

The preview images of the testing version are now the same size as the ones in the stable version. 
Before this change even if users tried to install the stable version it would try to load the preview images from `testing/live/Neko/images` and fail because they were too large.


Meisterlala:
An alternative to this would be to just remove the Neko Fans plugin entirely from the testing folder. But im not sure if that would break anything

philpax (approved):

PR #1303, opened 2023-02-10
DailyDuty v3.2.3.0

MidoriKami:
`Challenge Log` - Rewrote module, you can now select any challenge log entry for tracking
`Challenge Log` - No long includes chat notifications aside from "x allowances remaining"

`Duty Roulette Overlay` - Rewrote module, prevents text color from going wonky if updated when duty finder is open

`Wondrous Tails Overlay` - Fix minor memory leak (this never caused any problems, fixing it is a matter of pride)

`Wondrous Tails` - Prevent reroll message from showing when you have no second chance points

`Configuration` - Add automatic `{userid}.bak.json` backup configuration file in case the primary file becomes corrupted

### DIFF
--- a/stable/DailyDuty/manifest.toml
+++ b/stable/DailyDuty/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/DailyDuty.git"
-commit = "b04f503919a3b774b0b4f6730d94ed52c2785141"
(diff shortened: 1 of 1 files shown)


MidoriKami:
This is mostly a release from testing, 
```
Wondrous Tails - Prevent reroll message from showing when you have no second chance points
```
This is the one modification that wasn't in the current testing build.

philpax (approved):

PR #1336, opened 2023-02-16
[Testing] Allagan Tools v1.3.0.6

Critical-Impact:
Don't include lumina dlls with plugin


Critical-Impact:
Very minor change, just stopping lumina.dll from being output which is definitely causing issues.

goaaats (approved):

PR #1338, opened 2023-02-16
[Testing] CameraLoader

NeNeppie:
Lets you save unlimited camera configurations/presets in gpose at a time, vastly improving over the vanilla feature which lets you save only one. This makes taking quality screenshots easier and more consistent.

To prevent weird camera angles only achievable through gpose in normal gameplay, and likely crashes (I haven't tested what actually happens), the plugin is only useable within gpose. At least it should be.

Settings saved:
* Distance from the player
* Horizontal rotation angle (In relation to the player)
* Vertical rotation angle
* FoV
* Roll angle
* Pan & Tilt (Mentioned on the blue instructions window found in GPose)


philpax:
bleatbot, approve

philpax:

Infiziert90:
Tested your plugin for the PAC as version 1.1.0 and noticed a bigger issue and some smaller ones.

Major Issue:
- Doesn't work if a vanilla setting is disabled, System -> Other Settings -> Switch online status to Camera Mode while active
  + This setting completely disables the camera status, which leads to one of your checks failing
  ` bool isInCameraMode = Service.ClientState.LocalPlayer?.OnlineStatus.Id == 18;`

Major/Minor?
- Config values aren't validated, allowing a user to edit the config files with crazy values, load them as preset, and leave gpose with the crazy values still applied

Minor Issue: 
- Loading presets, while in ego perspective, breaks the position

- Deleting a preset makes the last created name index be -1, so a newly created preset gets an already existing name, leading to broken load behaviour
  + Example: 
  + You have Preset 8 and 9, and create a new Preset 10
  + Now you delete Preset 9
  + Creating a new preset will create it as "Preset 10", colliding with the already existing Preset 10
  + If you now try to load one of the presets 10, will it instead load the other preset 10

image

AzureGem also tested your plugin and found this minor issues:
+ Renaming a preset closes said preset. This is an ImGui ID issue, preset information is not lost or changed.

NeNeppie:
> Config values aren't validated

I was told that users externally modifying values is not an issue and is the user's responsibility. In retrospect, I may have misunderstood things. Either way, I have taken note of the settings that persist outside of gpose and I will validate those so they stay within the in-game parameters. 

> Deleting a preset makes the last created name index be -1, so a newly created preset gets an already existing name, leading to broken load behaviour

This I wasn't able to replicate. Presets in ImGui have a unique identifier based on their index, so when a preset is removed, the IDs of other/future presets change accordingly.
```cs
for (int i = 0; i < Service.Config.numOfPresets; i++)
{
            var preset = Service.Config.presets[i];
            
            if (ImGui.TreeNode($"{preset.name}##{i}"))
            {
```

As for the other issues, I have fixes either ready to go or in the works. I'll commit as soon as I'm able to.
Thank you for your time!

lmcintyre:
https://user-images.githubusercontent.com/6005409/230179723-7a0bf937-1913-4d92-aaed-1497acc8adbf.mp4
The above bug mentioned by Infi where preset names are duplicated still exists. It doesn't seem to break anything, but this should be fixed.
The bug is [here](https://github.com/NeNeppie/CameraLoader/blob/cf0d0851e58bc8c343a077d36ef129b4b20d82fd/CameraLoader/PluginWindow.cs#L151), where you are using the number of presets to name the new preset despite the new preset's name possibly already being taken.

One thing that would not *need* to be fixed, but would be nice to have is an explanation of Character Position vs Camera Position modes.

lmcintyre:
Sorry to give you more work to do, @NeNeppie, but would you be able to swap your icon for an icon that does not use assets from the game? Thank you.

lmcintyre:
Awesome, thank you!

lmcintyre (approved):

PR #1348, opened 2023-02-18
[Testing] StanleyParableXiv 1.2.4.0

rekyuu:
- Updated killing spree voice lines to be 30% quieter during playback
- Implemented Duty Complete playback for guildhests
- Disabled Duty Start and Duty Complete playback for solo quest instances
- Prevented Duty Failure from playing during FC workshop craft cutscenes

### DIFF
--- a/testing/live/StanleyParableXiv/manifest.toml
+++ b/testing/live/StanleyParableXiv/manifest.toml
@@ -1,8 +1,11 @@
 [plugin]
 repository = "https://github.com/rekyuu/StanleyParableXiv.git"
-commit = "7b8fe0db4850935cedd5e22eb5271242fa02cbc9"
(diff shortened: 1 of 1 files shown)


philpax (approved):
Looks good, consider fixing the warnings

rekyuu:
Thanks. The warnings are due to NAudio for some reason requiring Windows targeted libraries, so in order to develop/test on Linux I have to add this line to change the build target to `net7.0` instead of `net7.0-windows` when built on a Linux machine:

https://github.com/rekyuu/StanleyParableXiv/blob/main/StanleyParableXiv/StanleyParableXiv.csproj#L15

When the plugin is built in a Windows environment, those errors should be non-existent.  I was under the assumption the build pipeline built within a Windows container but it seems to be Ubuntu instead. I'll probably look at that later to see if there's a better solution or not.

PR #1349, opened 2023-02-19
[Testing] AudibleCharacterStatus 1.0.1.3

Arrenton:
Moving to D17, updated for API 8


MidoriKami:
>> Specify a project or solution file. The current working directory does not contain a project or solution file.
```toml
project_path = "AudibleCharacterStatus"
```
this should be:
```toml
project_path = ""
```
Since your .csproj is in the git root.

philpax (approved):

PR #1356, opened 2023-02-20
Allagan Tools v1.3.1.0

Critical-Impact:
Crafting calculation fixes
Reworked tooltips(new implementation + more display options) Character/retainer world is now tracked + source world filter Added wildcard searching
Added IPC service for getting item counts, enabling/disabling filters, managing craft lists and item add/remove events Fixed an issue with the class job filter


Critical-Impact:
Same version that is on testing if that helps in any way.

karashiiro (approved):

PR #1359, opened 2023-02-21
DailyDuty v3.2.3.3

MidoriKami:
Minor UI rework
- Removed Floating Info Button
- Replaced labeled buttons with icon buttons, included tooltips
- Added Overlay Configuration window that has both Timers and Todo overlay configuration
- Added Timer Style Window that has all timers in one window
- Removed outdated text from About window


MidoriKami:
image

karashiiro (approved):

PR #1361, opened 2023-02-21
[Testing] Hunty 0.6.0

Infiziert90:
I already submit this to the repo as the size is a bit bigger. 
Most of the changes come from my json rework, the true diff should be around 800ish additions and deletions when excluding json stuff.

+ add memory tracking (thanks Midori)
+ remove old tracking completely
+ redo monster.json for new tracking
+ open duty finder on location click
+ react to login/relog
+ prevent null exception with no gc
+ cleanup

- Support all global languages
- UI localization, currently supported languages en, de, fr, jp
- Add small indicator if rank / area is done
- More location updates

### DIFF
--- a/testing/live/Hunty/manifest.toml
+++ b/testing/live/Hunty/manifest.toml
@@ -1,11 +1,22 @@
 [plugin]
 repository = "https://github.com/Infiziert90/Hunty.git"
-commit = "d8161c5524c0dfb66c0ffe301f67bb51bd158d52"
+commit = "2ca8f3dff1490eb21fc5a1ce036e62ec1858a550"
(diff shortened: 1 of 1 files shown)


MidoriKami:
Specifically require 2856060e936f23b074635b36197c11e583432cdc or Newer.
Feb13 CS commit

Infiziert90:
bleatbot rebuild

Infiziert90:
bleatbot, rebuild

Infiziert90:
Can be merged now, stable got the CS update

reiichi001 (approved):

PR #1371, opened 2023-02-23
[Testing] BetterMountRoulette 1.1.0.8

CMDRNuffin:
Support for multiple characters!

Also adds a feature to prefer multi-seated mounts while in a party.


CMDRNuffin:
Will reopen in a bit/make new pr when I have figured out why select new mounts on unlock is borked.

karashiiro (approved):

PR #1380, opened 2023-02-26
[Testing] Retainer Track 1.0

carvelli:
Keeps track of who a retainer belongs to, in a vein similar to PlayerTrack - if you have seen the retainer's owner.

This is matched via the player's `ContentId`, which is typically available if:

- you are in the same party/alliance
- have seen an item crafted by them (where the crafter signature is available, and doesn't just say `Obtaining Signature`)

The simplest way to find retainer names is typically by hovering over crafted items - for a good chunk of players, they sell their own items. The plugin looks only at the owner's id - if you are reselling someone else's item, it won't accidentally detect the wrong player name. 

If found, the retainer's name is formatted as `Player Name (Retainer Name)`, otherwise it is left untouched. 

Example, with names partially blurred, if you see a tooltip such as:

image

This would then show up in the market board as follows (where the player has two retainers, one starting with `E`, one starting with `V`)

image


reiichi001:

PR #1385, opened 2023-02-26
Adopting WaymarkPresetPlugin

squidmade:
nofranz


squidmade:
I put nofranz as the description in case that’s needed to avoid a message in the plugin updates channel, but I would like to adopt this plugin since the original author put it up for adoption.

squidmade:
@PunishedPineapple has given the go ahead and offered to answer questions as needed https://github.com/PunishedPineapple/WaymarkPresetPlugin/issues/36#issuecomment-1445754766

reiichi001:
Adoption looks good to me. The `nofranz` should prevent this from popping up in the plugin release channel.

reiichi001 (approved):

PR #1397, opened 2023-03-01
[Testing] UsedName v0.8.1.0

LittleNightmare:
- Use memory instead of network packages for updates
- Add a tiny control window, use '/pname main' to open it


LittleNightmare:
here is the [diff](https://github.com/LittleNightmare/UsedName/compare/091de0a22b9366dd3b31f25df4d5ff97bf75d95c..ec3a30ffb20807c87dd8e09da1cabc8a4c606acf) compare with last pr in stable

PR #1423, opened 2023-03-06
Congratulations 1.2.0.0

Berna-L:
Get aurally rewarded when you get commended on duties.

Comes with default sounds for when you get commended by 1/3, 2/3, 3/3 or all seven members of the duty you were in, but you can set custom sounds for each type of trigger by opening /congratsconfig.

The fractioned values are in relation to how many of those members could commend you (i.e. excluding those that were in your party before joining the duty).


Berna-L:
This is the first release of the plugin on stable.

reiichi001 (approved):

PR #1445, opened 2023-03-08
Hybrid camera update to 1.1.0

Drahsid:
This now works with the most recent update. To do this, I borrowed the GameConfig code from SimpleTweaks. On top of this, I added tooltips, and a minimum size of the config window. Furthermore, I added some suggestions which the user may want to consider to improve their experience.


reiichi001:
@Drahsid for future pull requests, please make sure your DalamudPluginsD17 fork/branch are up to date with the current main brain here. That will prevent the superfluous commit history showing up.

reiichi001 (approved):

Drahsid:
Oh you're right. I was already syncing it, but I probably need to do that more aggressively.

PR #1452, opened 2023-03-08
[Testing] Chat Extended

Haplo064:


testing/live/ChatExtended/manifest.toml
```diff
@@ -0,0 +1,9 @@
+[plugin]
+repository = "https://github.com/Haplo064/ChatExtended.git"
+commit = "05a06e575edaeec9a026609b49dba70c64b3e605"
+version = "0.0.0.1"
+owners = ["Haplo"]
```
reiichi001, line 5: @Haplo064 this should be your github username

Haplo064:
This plugin is a revisit of my original plugin:
https://github.com/Haplo064/ffxiv-chat-extender

Since then, many changes have occured within Dalamud, and also my skills.
Currently, the list of abilities in the plugin include:
- Being able to change the time of anouncment messages, from GMT to your local timezone
- Logging of all chat channels to customisable files

Features that will be included in the future:
- Infinite chat channel groupings
- Better /tell handling
- RP features, such as highlighting, colourisation, noises
- Searching the chat
- Distance based messaging
- Commands for swapping chat group channels
- And many more as they either come to me, are requested, or are part of other requests in the request git.

All of this will be within the native chatbox, rather than an imgui window, as I feel this detracts from the user experience.
All within the rules for no cheating etc, as all of my plugins are.

lmcintyre:
Major
- The plugin throws an exception on every chat message for me. The exception is an IndexOutOfRangeException on line 42 in ChatLogger.cs.
- The plugin's config succumbs to the same issue you've had before iirc, if enums shift then every user's config will shift.
- Time localizing does not work for me. The announcement timestamp stays in PDT.

Minor
- It is possible to name two chat log files the same thing. It will write twice to each log if the same channels are enabled. Not sure if this is intended. I can see it being useful where one could have different formats for different channels within the same log file. However, if that's a supported use-case it's not a very user-friendly flow, but it's still a nice option to have.

Haplo064:
Major
* Throwing an error on every chat message, I'll fix up. That's a definitely shouldn't be happening (and I've probably fixed it locally already since then ,as I don't see any)
* The localisation needs more messages, then. It worked just fine for the few I tested with, but depending on if they change their message structure it won't work (This is why I want it in testing!)
* With regards to the enums as config, this I believe is the best method (that I can come up with) that won't break if changes are made to the enums (as seen with chat bubbles). I'm happy to have it shot down, if that's a requirement, and I'll have a big think about how I can have it be static, and yet also fluid if changes need to occur in the future.

Minor
* I didn't consider two chat logs having the same name. I do like the idea that it allows different formats in the same file, but I'll have to tweak the config to explain it a bit more to alleviate the user-unfriendly-ness of it.

Of note, is that I currently have a major addition in the works, and I won't be making any changes to these base functions in this branch until I've finished that one. Would it be better to remove this merge request, and request a new one in future (referencing this), or leave this on hold until I've finalised these changes and ping that that's done?

lmcintyre:
> With regards to the enums as config, this I believe is the best method (that I can come up with) that won't break if changes are made to the enums (as seen with chat bubbles). I'm happy to have it shot down, if that's a requirement, and I'll have a big think about how I can have it be static, and yet also fluid if changes need to occur in the future.
I've opened a PR in your repo pertaining to this, as it's far easier than explaining: https://github.com/Haplo064/ChatExtended/pull/1

Note that when I began testing both the base and my PR'd versions, the exception on each message went away. Not sure what the deal was.

Finally, if you're intending to find a significant number of your bugs when the plugin is in testing (which is completely fine, as there are no client crashes!), then I would recommend closing this until you're ready with your major changes, as maintaining small fixes to the testing version while you have another version in the works is not an enjoyable situation to be in.

philpax:
Heya! Have you had a chance to fix this up? If not, we might close it and let you reopen it at a later stage as 6.4's coming up soon and we'd like to have a clean slate for that.

philpax:
I'm going to close this as it has not seen any activity in a while and there have been several game updates since. Feel free to open a new PR to bring it back up to date if you'd like!

PR #1457, opened 2023-03-08
[Testing] Linkpearl 1.0.0.0

NotNite:
nofranz

An actually updated Mumble positional audio plugin. Because people still use Mumble... right?

This plugin allows you to hear your friends in 3D space in FFXIV, like their voice was coming out of their character in game, and your ears were on your camera.

Requires you to enable "Link to Game and Transmit Position" and the "Link" plugin in your Mumble settings. **Do not enable the Final Fantasy XIV plugin** - it's designed for a very old version of FFXIV and is not associated with this plugin.

Plugin reviewers with an NA/EU character can contact me (`NotNite#0001`) or Rika (`rin#3487`) for a demo on my Mumble server.

Technical details:
- Uses hashed content ID to identify players
- Uses a combination of world ID and territory type to assign context
  - When in duty, the world ID is replaced with `duty`
- Does *not* check if you're in a party
  - Maybe will add an option for this in a future update


NotNite:
The commit I just pushed adds my .editorconfig and resolves the security concern from sending non-hashed content IDs.

lmcintyre (approved):

PR #1475, opened 2023-03-10
[Testing] CollectableCalculator 1.0

carvelli:
Shows how many scrips and items you would receive if you turn in all items currently in your inventory, taking into account the collectability(quality) of items.

For example:

image

Would end up like this once turned in at Collectable Appraiser:

image


lmcintyre:
Hi! Please change your plugin icon to use a non-game asset. Thanks.

lmcintyre (approved):

PR #1479, opened 2023-03-10
[Testing] Who's Talking 0.1.0.0

sersorrel:
Draws voice activity indicators on the party list when you're in a Discord voice call to show who's talking/muted/deafened.

image

Known issues:

- There's _apparently_ a slow memory leak in the websocket library I'm using, which leaks ~200 bytes per reconnect attempt; in this plugin, that will happen once every minute if Discord isn't running, or once every five minutes if no messages are received from the Discord client (i.e. no voice activity and nobody joining/leaving the current voice channel). I suspect this memory is released when the connection next succeeds, but I'm not certain. If the game were to be left open for 24 hours with Discord closed, leaking 200 bytes per minute the whole time, this works out to less than 300KB leaked.
- The Discord API which exposes voice activity information is restricted to a small set of client IDs, and new applications are not currently accepted. This plugin currently uses the client ID used by [a reimplementation of the Discord overlay for Linux](https://github.com/trigg/Discover).
- Weird stuff might happen if two Discord users or XIV characters have the same name. (this seems kind of unavoidable...)

If any of those are blockers for getting this into testing, let me know, but it's worked fine in all my testing so far.


philpax:

lmcintyre:
Hi, sorry for the delay. @sersorrel I noticed you've updated the plugin since your last submission. Would you like to update it again, or do you want this reviewed based on 0.1.0.0?

sersorrel:
The updates are super minor, but I won't have time to bump the version number and test in-game until next week, so go ahead and review this version – I'll submit a separate update at some point, hopefully with some more features to make it a little more worthwhile

lmcintyre (approved):

PR #1523, opened 2023-03-22
[Testing] DailyDuty v4.0.0.0

MidoriKami:
Completely rebuilt plugin. Lots of new features, like a lot, you'll just have to install it to see =D

Biggest new feature, the Todo list is now a native ui element, meaning it draws on the same layer as the game.
image

### DIFF
--- a/testing/live/DailyDuty/manifest.toml
+++ b/testing/live/DailyDuty/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/DailyDuty.git"
-commit = "12f52e156e2cb563faf7cc8709dfe75845712959"
+commit = "dd59c9c84656caf4dcdba3f92d543cee9c1d835f"
 owners = ["MidoriKami"]
 project_path = "DailyDuty"
--- a/testing/live/DailyDuty/manifest.toml
+++ b/testing/live/DailyDuty/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/DailyDuty.git"
-commit = "06266bdde95c21cf9e87606a9ff1e5c3f1c10a8a"
+commit = "da74a6fa194802c6c40f235fc8a569c796cb7b53"
 owners = ["MidoriKami"]
 project_path = "DailyDuty"
--- a/testing/live/DailyDuty/manifest.toml
+++ b/testing/live/DailyDuty/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/DailyDuty.git"
-commit = "dd59c9c84656caf4dcdba3f92d543cee9c1d835f"
+commit = "48498cc5f437b548f47209f5e85362dbcf4d8148"
 owners = ["MidoriKami"]
 project_path = "DailyDuty"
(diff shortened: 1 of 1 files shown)


MidoriKami:
This version is a from-the-ground-up rewrite of DailyDuty. I do not recommend using the diff's to see what changed... since it's everything. This version actually has significantly less code.

Notes about the new design:

#### New Configuration System
A folder is created for each character that uses DailyDuty, inside this folder will be configuration and data files for each module.
Each module will have a `{modulename}.config.json` and a `{modulename}.data.json`

To prevent access exceptions, modules set a `datachanged` or `configchanged` module variable to trigger a single file write at the end of the update cycle, this limits writing to once per frame.

Configuration is entirely separate from data now, so users are free to copy config files between characters!

#### New Tracking System
All modules track their associated data at all times, most of the tracking is done via ClientStructs.
This is done so if a user decides to track something that they had disabled earlier in the week, the relevant data is already saved.

DailyDuty uses a single OnFrameworkUpdate function to update all modules, estimated max average load is approx 0.1000ms/frame

#### New UI System
Inspired by SimpleTweak's `Auto Config` option, I have implemented a UI drawing system using reflection to draw the appropriate imgui UI element depending on datatype.

This does have a performance cost, however it seems to cost significantly less performance than my old InfoBox elements that store draw commands as Actions in a list.

Estimated max average load is approx 
0.3500ms/frame when viewing data tab with a module selected
0.1500ms/frame when viewing configuration tabs with a module selected

#### Patch Resilliance
~~stealing~~ Borrowing techniques from SimpleTweaks, this version hooks the AtkManager's functions that create and finalize addons allowing DailyDuty to get Setup and Finalize events for any addons it needs with just 2 sigs.

The third and final sig in DailyDuty is for GoldSaucer Network Messages.

#### Abstract Module
Most of the modules logic is pre-defined in the Abstract Module class, this is inherited to implement DailyModule, WeeklyModule, and SpecialModule.

This makes each individual module extremely simple

#### Send Chat
The previous version of daily used send chat to send `/partyfinder` to open the party finder, this version instead checks if the quest `Pirates Maybe` is complete, and the user is not currently in a duty, before invoking AgentLookingForGroup.Show

#### Removed Features
`Timers` have been removed, it's not a feature I ever used, and because of that was often buggy
`Duty Roulette Duty Finder Overlay` has been removed and implemented as a SimpleTweak
`Wondrous Tails Duty Finder Overlay` has been removed, it has not been implemented as a SimpleTweak

MidoriKami:
`Tribal Quests` Removed Unused HelpText
`BaseModule` Fixed GetIncompleteCount logic error for Raid modules

MidoriKami:
Added missing PvP Checks

DailyDuty will deactivate completely when in any PvP area including the Wolves Den

MidoriKami:
- `Wondrous Tails` Add Forgot Book Warning when you walk away from Khloe without a Wondrous Tails book.
- `Tribal Quests` Fix using wrong localization key
- Use ExecuteMainCommand instead of AgentModule->LookingForGroup.Show
- Add Single Precision floating point view
- `Grand Company` Fix time displaying 17:59:59 when mission is complete

MidoriKami:
`Wondrous Tails` Add `Teleport` and `Return` to Forgot Book Warning

karashiiro (approved):

PR #1581, opened 2023-04-02
[Stable-New] AudibleCharacterStatus v1.0.1.3

Arrenton:
Updated for API 8. Changed the way the volume slider works and adjusts the volume of audio much better.


stable/AudibleCharacterStatus/manifest.toml
```diff
@@ -0,0 +1,10 @@
+[plugin]
+repository = "https://github.com/Arrenton/AudibleCharacterStatus.git"
+commit = "d4615cb10e8c6960f29609b595f219e7374245c9"
+owners = [
+    "Arrenton",
+]
+project_path = "AudibleCharacterStatus"
```
MidoriKami, line 7: Remove this line, your .csproj is in the root, this is only needed if the project folder is nested.

reiichi001 (approved):

PR #1597, opened 2023-04-04
Hit it, Joe! (Formerly TF2-ish Critical Hits Plugin) 3.0.4.0

Berna-L:
New name: **Hit it, Joe!**
- Since the plugin hit version 3, keeping the name Valve-related would cause crashes.

[Countdown Jams]
- New module.
- Set sounds to be played when a countdown starts.

[TF2-ish Critical Hits]
- New algorithm to differentiate your heals from other players'. (Thanks, Aireil!)
  - Also considers heals from your fairy as yours.


Berna-L:
Commit hash the same as https://github.com/goatcorp/DalamudPluginsD17/pull/1585.

reiichi001 (approved):

PR #1708, opened 2023-04-20
[Testing] Allagan Tools v1.5.0.0

Critical-Impact:
**House Storage has arrived!**
So this took a while but it has finally come to fruition. A few things to note:

- To have a house register with the plugin you must first enter it, have permission and then open the 'Indoor Furnishings' menu. This will allow for the plugin to see you own the house and add it to your 'Characters' list.
- Once the house is registered due to the way the inventory data of each section is provided, you must enter each section to have it be parsed by the plugin. For Indoor and Outdoor Furnishings you must enter the storeroom tab before that data is collected.
- For Interior Fixtures open the relevant section in the housing menu.
- There's a lot of moving parts so if you run into issues, bugs or crashes hit up the #plugin-help-forum on discord.
- I'll be working on making the 'Is Housing Item' filter a bit more reliable as this might be more important now.

### DIFF
--- a/testing/live/InventoryTools/manifest.toml
+++ b/testing/live/InventoryTools/manifest.toml
@@ -1,9 +1,18 @@
 [plugin]
 repository = "https://github.com/Critical-Impact/InventoryTools.git"
-commit = "1b68b29f746dbf54a3d24442c829f60f55485414"
+commit = "f677c3a5445cfc473c7f1d74f24c81cadea10fb1"
 owners = [
     "Critical-Impact",
 ]
 project_path = "InventoryTools"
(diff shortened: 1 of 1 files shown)


Critical-Impact:
The live version is more up to date than the testing version so the diff is kind of off
The better diff would be https://github.com/Critical-Impact/InventoryTools/compare/3dfd116989bfb070ef993a6b1c842fb70021f17b..f677c3a5445cfc473c7f1d74f24c81cadea10fb1

As that is comparing the latest commit on main with the last commit that was released on live.

reiichi001 (approved):

PR #1711, opened 2023-04-20
Better Shadows 1.1.4 (Performance improvements)

Drahsid:
Out of morbid curiosity, I checked the plugin statistics and to my horror saw that Better Shadows was by far at the top. Of course, the reason was pretty obvious. I was sorting the zone configs every frame lol. Now these things only happen when you change zone, which should be more reasonable. The improvement on average draw time is over 1000x lol


goaaats:
bleatbot, rebuild

reiichi001 (approved):

PR #1726, opened 2023-04-23
EmoteLog 1.0.1.0

RokasKil:
- Added font size configuration
- Added text wrapping configuration


goaaats (approved):

goaaats:
FYI we don't guarantee that those font file names are static. Might be better using the glyph request APIs.

PR #1733, opened 2023-04-26
Team Fortress Fantasy 1.1.1.0

Berna-L:
[Win Panel]
- Added option to have the Win Panel save the score per duty.
  - This is the default behavior for new installations.
  - Current users will be told about this through chat when updating the plugin.
- Added window (accessible in the Win Panel configuration) to check the saved scores per duty.
  - This window also has an option to copy the values as CSV to the clipboard and delete individual scores.
- Fixed the MVP list closing when pressing ESC.


Berna-L:
bleatbot, rebuild

goaaats (approved):

PR #1742, opened 2023-04-28
[Testing] RPBlurb v1.0.0.1: bugfix: ImGui text centering style spillover

meoiswa:
nested returns bad.


goaaats:
FYI, you shouldn't version your changelog. Plogon already does it for you. Like this, it's duplicated every time you make a release.

goaaats (approved):

PR #1788, opened 2023-05-12
Update repo locations

carvelli:


carvelli:
I don't know why the build won't work, but I admittedly don't care very much either; I won't be back to check on the status of this PR.

reiichi001:
From my understanding, there's nothing to build since the git hashes aren't changing. This could be considered a plogon bug, since adding ownership doesn't necessarily need to trigger a build.

@mabako can fix this by doing a version bump and updating the commit hashes.

reiichi001 (approved):

PR #1797, opened 2023-05-15
VFXEditor 1.7.5.0

0ceal0t:
- Large UI refactor
- Reworked documents and renaming
  - Documents displayed as a tab bar rather than in a separate dialog
  - Documents can now be renamed by right-clicking the tab
  - Favorited selection items can be renamed by right-clicking
  - Nodes in the VFX library can be right-clicked to import/delete/rename
- TMB tweaks


philpax (approved):
I hope you don't need to do any large UI refactors again any time soon 💀

PR #1816, opened 2023-05-22
PartyPlanner 1.5.2

edg-l:
- Fixes an issue where some events were not shown.


MidoriKami:
You don't need to PR the same version to both stable and testing-live.

If stable has a higher version number than testing, users will get the stable version, even if they have testing enabled.

edg-l:
> You don't need to PR the same version to both stable and testing-live.
> 
> If stable has a higher version number than testing, users will get the stable version, even if they have testing enabled.

i see, i didnt know that, do i need to change the pr now however?

reiichi001:
@edg-l can you verify this is good for 6.4 please? If so, we'll merge it in.

edg-l:
@reiichi001 tested and works

reiichi001 (approved):

PR #1888, opened 2023-05-26
[Testing] A Realm Recorded v0.7.0.0

UnknownX7:


UnknownX7:
Probably will be 1.0.0.0 as well if no major bugs are found. This update doesn't fix any major bugs either so take however long is needed.

reiichi001 (approved):

PR #1910, opened 2023-05-27
 [Testing] Hit it, Joe! (Formerly TF2-ish Critical Hits Plugin) 3.4.0.0

Berna-L:
[TF2-ish Critical Hits]
- Add option per job to set a minimum time between sounds.
  - For example, if you set the time as 1000 ms and a critical hit sound is played,
  no other sound (be it for critical hit, direct hit, critical heal etc.) will be played for the next second.
  - Keep it at 0 ms to use the current behavior.

(Thanks to Grayve for the idea and everyone at the Discord thread for the feedback!)


Berna-L:
Diff between 3.3.2.0 in Stable (#1828 ) and this one: https://github.com/Berna-L/ffxiv-tf2-crit-plugin/compare/54ee390e6dc42a11512edee6d24d132e2e973ff6..04c67c409ccbcba45e1a1342d036482521cc32f2

karashiiro (approved):

PR #1964, opened 2023-06-04
MemoryMarker v1.0.0.5

MidoriKami:
Fix only showing first word in rename window

Added support for SeString symbols, now you can add symbols to your names

image


MidoriKami:
Realized the moment I looked at the diff that I forgot a dispose, should be good after it builds.

reiichi001 (approved):

PR #1970, opened 2023-06-05
Allagan Tools v1.5.0.6

Critical-Impact:
Actually fix the housing crash, much appreciated to Laissabelle for helping me track it down


Critical-Impact:
Apologies for the double release in one night, worked through with a tester and I believe I've narrowed down the real bug.

reiichi001 (approved):

PR #2000, opened 2023-06-11
SubmarineTracker 1.6.0.0

Infiziert90:
[Info]
This update changes the internal loot tracking significantly and
old loot data is now marked as legacy.

[Changes]
+ Loot tracking update
+ Internal improvements
+ Voyage history includes more information now, like procs

[Important]
All config files must be migrated to the new version, and 
this is only possible while 1 client is running.
Running multiple clients will result in a plugin crash
to prevent any corruption of these configs.

### DIFF
--- a/stable/SubmarineTracker/manifest.toml
+++ b/stable/SubmarineTracker/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Infiziert90/SubmarineTracker.git"
-commit = "1d1b484e4e36d0bf92841e81cb9007226947f266"
+commit = "b7dc8c8ad7e219b70f1f98b103f896156f4362b6"
 owners = [
     "Infiziert90",
 ]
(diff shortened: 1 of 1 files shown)


karashiiro:
While I think multiboxing is uncommon, could you make this throw an exception rather than destroy config in the case where someone is multiboxing? I think most people see changelogs after updating rather than before updating, and don't check the updates channel first.

karashiiro (changes requested):
See comment

Infiziert90:
I already thought about a way of making it throw an error, but haven't really come up with any good solution

karashiiro:
I don't think it needs to be handled gracefully, so long as it breaks before destroying config rather than after. I think checking for multiple `ffxiv_dx11.exe` processes at the beginning of the migration and throwing if so is good enough.

karashiiro (approved):

PR #2033, opened 2023-06-23
[Testing] Mappy v1.0.0.0

MidoriKami:
Mappy, Returns with a Vengeance.

This plugin is a complete replacement for the vanilla game map.

**Mappy will not work exactly like the vanilla map, don't expect it to.**
Don't bother posting anything along the lines of "well vanilla does it this way"

This plugin is still very much a **work in progress**.
Many things will change with this plugin while it is in testing.
Bugs are to be expected. You are testers, so go test!

Please Post ALL issues to the github <https://github.com/MidoriKami/Mappy/issues>


MidoriKami:
This is a complete rewrite of the plugin, there have been no new main features added to this plugin.

Performance is significantly better than previous versions.

karashiiro (approved):

PR #2040, opened 2023-06-27
[Testing] Mappy v1.0.0.1

MidoriKami:
Improved tooltips, they now have icons and look really cool
Improved plugin defaults, each module now has a reasonable default layer

Added option to display tooltips for misc map markers, ie, "Shop", "Repair", "Trader"
Added more tooltips for special icons

Treasure icons now no longer have a planar distance limit, instead they have a height limit, you can't see treasure more than 20y above or below you on the map.

**It is recommended to right click the plugin in the installer and click "Reset plugin configuration and reload"**
If you find think some of the defaults are a bad idea, let me know at the github issue tracker: <https://github.com/MidoriKami/Mappy/issues>

image

### DIFF
--- a/testing/live/Mappy/manifest.toml
+++ b/testing/live/Mappy/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/Mappy.git"
-commit = "1f84898458fa812b6bb8963029a0ca3ee4fa8964"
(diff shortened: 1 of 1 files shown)


MidoriKami:
Sorry for the huge PR right after it was merged, I had hoped to get this in before v1.0.0.0 was reviewed.

lmcintyre (approved):

PR #2057, opened 2023-07-01
[Testing] DailyDuty v4.0.3.0

MidoriKami:
Added OnClick actions to Todo List.


MidoriKami:
Most of this diff is splitting out config and data into their own files for each module.

Using interfaces to define the data driven UI would have been a bit too bulky to keep in the individual module files.

reiichi001 (approved):

PR #2069, opened 2023-07-02
Simple Tweaks 1.8.8.2

Caraxi:
***Tweak Changes***
- **`Expanded Currency Display`** - Fixed positioning of gil display moving when scale is anything other than 100%


Caraxi:
Change is bigger than it really should be because I commented out a whole new tweak that isn't being released this update

karashiiro (approved):

PR #2119, opened 2023-07-15
DeathRecap 1.10.2.2

Kouzukii:
Added Dismantle to the list of captured status effects (thanks Speedas)


philpax (approved):
You might want to use constants for these so that you don't have to look them up from the integers

PR #2124, opened 2023-07-16
Wholist v1.3.1.1

Blooym:
nofranz

Updating various links and names in the plugin to point to new username/domain


philpax (approved):
new username who dis

PR #2128, opened 2023-07-17
TitleRoulette 1.3

mabako:
- Picking a random title now makes sure it doesn't pick your current title
- "{no title}" added as an extra option to clear your current title


reiichi001:
bleatbot, rebuild

reiichi001:
bleatbot, approve

PR #2134, opened 2023-07-18
DeathRecap 1.10.2.3

Kouzukii:
Add BLU Bad Breath, Conked from Magic Hammer, and Candy Cane (new in 6.45) to the list of captures status effects. (Thanks xiashtra)


philpax (approved):
I hope you remember what these codes represent

PR #2136, opened 2023-07-19
[Testing] SortaKinda v0.0.0.1

MidoriKami:
New Plugin, SortaKinda.

This plugin allows you to define sorting rules and automatically sort your inventory according to these rules!

**This is a Testing Plugin, please expect bugs, issues, and crashes. You're testers, so go test!**


MidoriKami:
Code is now as clean as it's going to get 🙃 
Been testing the functionality personally for a while now and it seems to be working quite well.

Critical-Impact:
Tested this with Allagan Tools and Inventory Search Bar. Any plugin that relies on reading ITEMODR for the true location of the item will return the incorrect result until an actual change happens to the inventory. If this does go out I'll need to let users know they can't use both plugins at least not in a reliable way

MidoriKami:
I am working on a bit of a re-design with the internal systems as things got to be a bit unmanageable at the end.
Marked this as a draft for now, the plugins feature set will remain unchanged from the last committed version.

I am able to somewhat address Critical-Impact's concerns by forcing an ItemODR write after sorting.

MidoriKami:
Significant code improvements.

MidoriKami:
Only KamiLib has been updated to use the new ITextureProvider in Dalamud, no changes in SortaKinda itself since last commit.
https://github.com/MidoriKami/SortaKinda/compare/d2f18f03c4ce2d41c2f52c6bfea887b63cf67283..e66434e8c829f1242bf31be20a5afcedf7d15e97

philpax (approved):

PR #2194, opened 2023-08-05
[Testing] PetRenamer v0.4.0

Glyceri:
NOW SUPPORTS BATTLE PET NAMES!!!!!!!!
+ Added support for renaming your Carbuncle and Faerie.
+ Added a new theme to indicate that you are now renaming a Battle Pet compared to a Minion.
+ Rewrote the naming process to be WAY more stable. Seeing others people's names should work better now.
If not. Please relog. After a relog names should always update.


Glyceri:
Why does it do this :(

Glyceri:
bleatbot, rebuild

PR #2201, opened 2023-08-06
[Testing] Who's Talking 0.4.0.0

sersorrel:
aaaaand of course someone spotted a problem (but not with the alliance raid stuff); it should now actually consistently follow your instructions if you told it which discord accounts match up with which xiv characters

also this adds a bonus list of "people not in your party who are speaking" just below the party list, with an option to turn it off in the plugin settings


philpax (approved):
love the change in text if you're using it from "testing"!

PR #2240, opened 2023-08-17
Mappy v1.0.2.3

MidoriKami:
Add `Hostiles` module that displays enemies.


MidoriKami:
PAC Notes: Only enemies that are in the object table are displayed, enemies that are "IsTargetable = false" are not shown, enemies that are "IsDead = true" are also not shown.

reiichi001 (approved):

PR #2241, opened 2023-08-17
XIVDeck 0.3.15

KazWolfe:
Is the space within a box in a box also part of the space of the bigger box? It can't be, because the space inside the box inside the box cannot be used to store things inside the first box. But it is still volume that can be used, so perhaps it is? This multi-box thing hurts my head.

- XIVDeck will now respect users who ask to not be alerted on multiboxing, thank you very much. The plugin will still not work with multiple clients, sorry.

For those of you not on the testing track, you're also getting all the [version 0.3.14 changes](https://github.com/KazWolfe/XIVDeck/releases/tag/v0.3.14) for free!

Full release notes, testing notes, and downloads are available [on the plugin GitHub](https://github.com/KazWolfe/XIVDeck/releases/tag/v0.3.15).

### DIFF
--- a/stable/XIVDeck.FFXIVPlugin/manifest.toml
+++ b/stable/XIVDeck.FFXIVPlugin/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/KazWolfe/XIVDeck.git"
-commit = "203abb50688eaa4c0af51b0ff0cb0ef8ce336517"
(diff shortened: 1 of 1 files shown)


reiichi001:
Easier PAC review link (0.3.14's release, which is in testing already) https://github.com/KazWolfe/XIVDeck/compare/d1e264b14374401703546898cfdb01223b4b8011..b7774a11a7aa6a0b975839c65e8036b170dc4774

reiichi001 (approved):

PR #2251, opened 2023-08-21
[Testing] DelvCD Initial Release

Tischel:
## DelvCD

The purpose of the plugin is to provide customizable UI elements that can appear or change given certain conditions.

**List of types of conditions you can set for your UI elements:**

### Cooldowns
* Remaining time 
* Charge Count
* Is Combo Ready
* Is Action Usable
* Is Target in Range 
* Is Target in Los

### Status Effect
* Is Active
* Remaining time
* Stack Count

### Character State
* Level
* HP
* MP / CP / GP
* Is Pet Summoned

### Item Cooldown
* Remaining time

### Job Gauge
* This is not done yet but its in the plans. It'll let you create conditions based on the job gauge data.

All these conditions can be combined. For example you could have an icon that only shows when your HP is below a certain threshold, a certain action is off cooldown and your level is above 50.

The conditions can determine if an element is visible or how it looks.
For example I could have an element showing my Lucid Dreaming cooldown, and have a conditions that adds an animation to it if my MP falls down a certain threshold.

### Visual Examples
**Custom Summoner hotbars**
https://i.gyazo.com/d37e688026812716743fd49eb28b3a60.gif

**Custom Machinist hotbars** (combined with DelvUI's Machinist HUD)
In this example you can see a square changing colors based on my 1-2-3 combo state (through DelvCD)
https://gyazo.com/152bc958581f7f7435c76d22f151226c

### Things this plugin can't do (and never will)
* Make conditions based on the state of other players / clients
* Make conditions based on fight mechanics
* Make conditions based on combat time / length
* Make conditions based on the combat log
* Provide any sort of automation for any sort of action

## Context

This is plugin is continuing the work by [Lichie](https://github.com/lichie567) with a rename.
We're planning on giving support on the DelvUI discord, hence the name.
For more context, we used to give support but that stopped a while ago when Lichie stopped updating/supporting the plugin.

I have Lichie's permission to do this and I can show our conversations if required.
Also note that the original plugin was never accepted in the main repo before, but there were some conversations in the Dalamud discord that led me to believe that it might be accepted now.

In the case this is rejected again I would like to know what specific changes can be made to get it accepted.

Also I've put this in the stable folder because even though the plugin is new here, it's been used to a long time. If you'd prefer for it to be added in testing first just let me know.


NotNite:
Above build failure is because the variable for where Dalamud libraries are located needs to be named DalamudLibPath.

Tischel:
> Above build failure is because the variable for where Dalamud libraries are located needs to be named DalamudLibPath.

Yep didn't realize the project file was so outdated, should work now, thanks

lmcintyre:
I think we'd like for this to be in testing, at least for a short period. We don't want anyone under the impression that plugins new to the main repo will be admitted without a testing phase, no matter how short.

Tischel:
> I think we'd like for this to be in testing, at least for a short period. We don't want anyone under the impression that plugins new to the main repo will be admitted without a testing phase, no matter how short.

Fair enough, will make the changes asap.

philpax (approved):

PR #2254, opened 2023-08-22
[Testing] Tick Tracker 0.0.1.0

Kurochi51:
- Reworked...everything?
- Tick progress is tracked more consistently
- Lists of Status IDs that can trigger faster ticks or halt regen altogether of mana / hp are now automatically generated
- Added a new option to hide the respective bars when hp or mp is full
- Disabled in PvP instances. It still works in Wolves' Den, but not while in the duel area.
- Added a debug window for ease of access to the generated lists


Kurochi51:
I heard PAC loves big updates, so I set out to rewrite the plugin as best I could. Also 682 lines from that diff are the [license](https://github.com/Kurochi51/TickTracker/commit/6ce5b31d07e32fae091db653dacbbed85768282a) change from MIT to AGPLv3

lmcintyre (approved):
Approved. Also, I hate you.

PR #2298, opened 2023-08-28
[Testing] HUDManager 2.5.17

nebel:
**New features**
- Added a "QoL Bar condition" custom condition type. If QoL Bar is installed, this allows you to use QoL Bar conditions for triggering HUD swaps.
- Added a "base classes" class/job group, which matches base classes but not jobs.
- Added the ability to re-order custom conditions.

**Fixes**
- Fixed a bug which caused changes made in a layout's Windows tab to be lost unless other settings were also changed.
- Fixed a bug which prevented a layout from being fully visible immediately after importing or creating a new layout.

**Changes**
- Modified the definition of "DoW" and "DoM" class/job groups to include the DoW/DoM base classes in addition to jobs. If you relied on the old behavior, you can instead create a custom multi-condition which excludes the new "base classes" class/job group.
- Made deletion of HUD elements and custom conditions require confirmation by holding the Control key.
- Made various small modifications to the UI to make the appearance of buttons, ordering of lists, naming of features and so on more consistent.

An addition to the above, a change was made to the way HUD swaps work in cases where a condition changes, but there are no actual changes to the layout or layers as a result of the changed condition. Previously the HUD was swapped in these cases, even though the previous state and the new state were the same. Now these swaps will be skipped instead. This is an experimental change to help reduce the performance impact of "unnecessary" HUD swaps. Let me know if you experience any issues related to this.

### DIFF
--- a/testing/live/HUDManager/manifest.toml
+++ b/testing/live/HUDManager/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/zacharied/FFXIV-Plugin-HudManager.git"
-commit = "454a9f67fd25a03cd12592d88b7944c1ce426486"
+commit = "a85881b406d802e5baf8425cf17e7d49898c2f62"
 owners = ["zacharied", "nebel"]
 project_path = "HUDManager"


goaaats:
bleatbot, approve

reiichi001 (approved):

nebel:
I'm the new maintainer/developer since Necoji no longer had time to maintain it. My name on Dalamud discord is nebel (username `settings.json`).

PR #2306, opened 2023-08-30
Mappy v1.0.2.7

MidoriKami:
Fix hostiles showing when viewing other maps


MidoriKami:
Oh. This is testing. Oops.

PR #2324, opened 2023-09-02
[Testing] Tick Tracker 0.0.1.2

Kurochi51:
- Fixed progress bars resetting erroneously when hp and mp regen is active
- Fixed progress bar resetting when directly healed


Kurochi51:
I moved the signature scan and hook init earlier in the ctor, to avoid needlessly initializing and disposing of extra things if the signature is wrong.

reiichi001 (approved):

PR #2338, opened 2023-09-05
[Testing] Tick Tracker 0.0.1.3

Kurochi51:
- Increased the level of log events to avoid unintended spam


Kurochi51:
Forgot to change these to verbose before release, also switched to imraii stuff where it made sense.

philpax (approved):

PR #2349, opened 2023-09-06
XIVDupeFinder to 1.0.1.0 (stable)

CallumCarmicheal:
Updating to latest commit, 1.0.1.0. Changing to stable (because I put it in the wrong folder, whoops).

- Fixed hooks to stop plugin running when disabled / closed.
- Fixed issue with highlighting always clearing when not enabled
- Minor optimizations and code refactoring
- Moving from Testing to Stable


CallumCarmicheal:
VS updated the incorrect version field, squashed the commits into 1 instead of 2.

reiichi001 (approved):

PR #2365, opened 2023-09-11
DelvCD release 0.4.0.0

Tischel:
- Conditions data options are now dynamically populated depending on the triggers in that element:
  * The condition properties were fixed before which resulted in confusing behavior if the element didn't contain the appropriate trigger types.
  * The "Dynamic" option was removed.

- Updated how the data from triggers is used for Icon swipe animations:
  * The trigger and its property must be explicitly set to determine what progress is being tracked.
  * Some of your Icons might need to be manually corrected.

- Updated how text tags work:
  * Text tags are now more explicit which provides more control on which trigger the label will use to get the data from.
  * For example to show the cooldown of an action, instead of `[value]`, you would now do `[cooldown_timer]`.
  * Existing text tags will be automatically changed when updating to this version.
  * Some of your Labels might need to be manually corrected.
  * Use the 'Tags' button next to the Text Format input field in Labels to see all the available text tags for each type of trigger.

- Added JobGauge triggers:
  * These allow you to use data from your job gauges to make conditions.
  * They only work when playing on the corresponding job.

- Added Bar elements:
  * Similar to the Icon's swipe animations, these need to be linked to a trigger property to track its progress.

- Added automatic settings backup system:
  * Every time DelvCD's is updated, a backup of your settings will be created in `%APPDATA%\XIVLauncher\pluginConfigs\DelvCD\Backups\<previous_version>`.
  
- Added `/dcd` command to open DelvCD's settings.

### DIFF
--- a/testing/live/DelvCD/manifest.toml
+++ b/testing/live/DelvCD/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/DelvUI/DelvCD.git"
-commit = "5652a7ed9d93961dae664910a53010a8a3f208e4"
+commit = "02261ec37be9c9eae1515624aceb4ceb24d2b811"
 owners = ["Tischel"]
 project_path = "DelvCD"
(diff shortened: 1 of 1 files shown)


philpax (approved):
I'd suggest making smaller changes in future 😅

PR #2382, opened 2023-09-15
[Testing] Tick Tracker 0.0.2.0

Kurochi51:
- Swap to a better system of tracking server ticks
- Use a more accurate method to update while regen is active
- Added support for Gathering Points


Kurochi51:
If there's an issue with `IPluginLog` being commented instead of straight up removed, let me know.

MidoriKami:
> If there's an issue with `IPluginLog` being commented instead of straight up removed, let me know.

Won't effect anything at all.

karashiiro (approved):

PR #2384, opened 2023-09-16
[Testing] Tick Tracker 0.0.2.1

Kurochi51:
- Fix edge case scenario where bars wouldn't update after a loading screen


Kurochi51:
169 of the differences are from `.editorconfig` and the github actions

Kurochi51:
Turns out my fix was an illusion, as things often are when multi-threading is involved. Otherwise same thing as before, 175 lines from the diff are just `.editorconfig` and github actions / readme.

karashiiro (approved):

PR #2387, opened 2023-09-17
Mappy v1.0.3.1

MidoriKami:
Add more Center on Open options. (from @pikajude)
Add option to disable centering on map links. (from @pikajude)


MidoriKami:
It seems the resource file was regenerated with a new format, that accounts for nearly 450 lines of the diff.

karashiiro (approved):

PR #2402, opened 2023-09-24
[Testing] DiademCalculator 0.0.0.1

Infiziert90:
Just a simple calculator for the current sessions diadem points and scrips

- Supports all 3 gathering jobs
- Automatically opens inside Foundation & Diadem
- Adjustable background opacity 

overlay


lmcintyre:
Hi! Sorry to disappoint but we have not gotten to this plugin yet, and will likely not be able to review it until after Dalamud API 9 has been released. I would recommend taking the time and swapping to the API 9 branch of Dalamud by using /xlbranch in-game and swapping to v9 so you can test compatibility with your plugin. Once 6.5 releases, you can verify functionality once more and submit a new build.

Thank you and apologies for the delay.

goaaats:
bleatbot, rebuild

Infiziert90:
Updated for API9

reiichi001:
Tested the current build for PAC purposes. When you lock the window, it becomes impossible to unlock. There is no way to interact with the settings after that.

Infiziert90:
> Tested the current build for PAC purposes. When you lock the window, it becomes impossible to unlock. There is no way to interact with the settings after that.

Same issue as with my other plugins, i used ### to scope windows without realizing that it would break things horrible

Is fixed in the new version

lmcintyre (approved):

PR #2411, opened 2023-09-29
[API9] SubmarineTracker 1.8.0.2

Infiziert90:
- Prevent crash in hook
- Add current sector unlocks
- Add predicted breakpoints
- Switch to the tracky track config save system, should improve stability


goaaats:
bleatbot, rebuild

reiichi001 (approved):

PR #2413, opened 2023-10-02
[API9] OBSPlugin 1.0.2.6

Bluefissure:
Change blur plugin to [obs-composite-blur](https://github.com/FiniteSingularity/obs-composite-blur) since StreamFX is behind a paywall.


goaaats:
bleatbot, rebuild

PR #2441, opened 2023-10-03
[API9] HarpHero v1.9

MgAl2O4:


reiichi001:
Code looks fine, but DalamudPackager will need to be bumped to 2.1.12 for API 9. After that, this will pass review

reiichi001 (approved):

PR #2447, opened 2023-10-03
Update all Anna plogons

lojewalo:
Update all my plogons


lojewalo:
This is ready after build. I tested that they all load and that their basic functions don't crash.

reiichi001 (approved):

PR #2463, opened 2023-10-03
Pixel Perfect 3.0.4.0

Haplo064:
Updated for API9
Added dashed rings option


reiichi001:
These go with the toml manifest here now instead of packaged with the plugin. Should be able to just copy the images folder wholesale and it'll work and pass plogon build checks. 

https://github.com/Haplo064/PixelPerfect/tree/master/images

Haplo064:
<3

reiichi001 (approved):

PR #2490, opened 2023-10-04
[Testing] Currency Tracker 1.2.3.4

AtmoOmen:
6.5 / API9
No functional changes.


reiichi001:
You've got an API override in your plugin manifest. You'll want to remove `"DalamudApiLevel": 8,` so that DalamudPackager takes care of it. Otherwise, this update will build, but will be auto-disabled for an old API level being displayed.

Other code is otherwise good, so this is the only blocker.

AtmoOmen:
Forgot that, it should work now

reiichi001 (approved):

PR #2528, opened 2023-10-04
PlayerTrack

kalilistic:
Remove from stable for now.


kalilistic:
Guessing fail expected since removed?

kalilistic:
This is to fix a bug where the testing version cannot be installed if the stable version has a lower API level. PlayerTrack can't be installed until this is merged. Goat confirmed it's okay here: https://discord.com/channels/581875019861328007/929533532185956352/1159233537640235090.

lmcintyre (approved):

PR #2534, opened 2023-10-05
[API9] TargetLines 1.3.1.0

Drahsid:
Pretty much another API9 update. Changed some minor behavior which should make occlusion more consistent, and besides that, possible optimizations in TargetLine.


Drahsid:
I bumped the version to 1.3.1, since I noticed and fixed a bug relating to the state switching of the lines. On top of that, I pretty much threw out my wrapper class and am now using an extension class on GameObject, which simplified some of the behavior (and with less `new`, seems to save a notable amount of time)

reiichi001 (approved):

PR #2536, opened 2023-10-05
EzWondrousTails v2.3.0.1

MidoriKami:
Fix clicking on duty tiles not opening the correct duties.

Performed partial plugin rework, new features planned in the near future.


MidoriKami:
Size is misleading, I changed the namespaces to File-Scoped namespaces, so it thinks every line is changed.

reiichi001 (approved):

PR #2560, opened 2023-10-05
Fungah 1.0.0.1

imaginary-png:
api9 update


reiichi001:
Please update DalamudPackager to 2.1.12 (or later) and remove any hardcoded API 9 entries from your plugin manifest.

Infiziert90:
Also please fix the icon, it throws an error in the installer because the conversion to png got corrupted.
Easy fix is to re-export it with any image tools like paint.net, photoshop

karashiiro (approved):

PR #2577, opened 2023-10-06
Distance 1.0.1.5

PunishedPineapple:
- Updated for API 9. Thank you Felys!

### DIFF
--- a/stable/Distance/manifest.toml
+++ b/stable/Distance/manifest.toml
@@ -1,15 +1,10 @@
 [plugin]
 repository = "https://github.com/PunishedPineapple/Distance.git"
-commit = "a28ed6cc29f61d56d91cad5a814fc54317ea8565"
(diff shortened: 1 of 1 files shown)


reiichi001:
For future updates, it's recommended to remove `"DalamudApiLevel": 9` from the plugin manifest since DalamudPackager will handle filling that in for you.

reiichi001 (approved):

PunishedPineapple:
> For future updates, it's recommended to remove `"DalamudApiLevel": 9` from the plugin manifest since DalamudPackager will handle filling that in for you.

Doesn't this run a bit counter to the point of having the API level specified?  At that point it basically just becomes a packager version check, not whether the plugin is intended to be a specific API level.  Generally, the other manifest fields that the packager inserts are taken from what your code states that it is (i.e., assemby version and internal name), and not assumed by the packager like the API level is (besides loading parameters, which I am guessing few people use, but maybe I am wrong).

PR #2580, opened 2023-10-06
Say What? 1.0.2.2

PunishedPineapple:
- Updated for Dalamud API 9.
- Minor improvement to settings window.
- Minor code cleanup.


PunishedPineapple:
The large line count for the diff is due to switching to file-scoped namespaces.  I am sorry for this.

reiichi001:
Code is reviewed and looks good aside from two things.

Please bump DalamudPackager to version 2.1.12 and remove the DalamudApiLevel field from the plugin manifest, as DP will handle that instead. This ensures that the generated plugin manifest has all fields and won't experience unforeseen overrides from mismatching with the build system.

PR #2623, opened 2023-10-07
Redirect 1.2.2.2

cairthenn:
- Fixes an issue that prevented sprint from activating


cairthenn:
The purpose of the changes is to prevent a bug with Sprint not being usable that emerged due to a change in ClientStructs.

The original PR for 1.2.2.1 has a hotfix which is functional, though uses some incorrect assumptions.

After some discussion as a result of the ClientStructs changes, I discerned I was incorrectly using `ActionType` in the plugin since the (original) values seemed to align with the few things I used it for. The follow-up PR correctly gets the `ActionCategory` from the game instead of using `ActionType`, and will also cause the plugin to do nothing with regards to Sprint and Potion queueing in the event that it cannot determine the game's original values.

reiichi001 (approved):

PR #2635, opened 2023-10-08
[Testing] Add MegaMappingway Uploader 1.1.0

lojewalo:
It's a live player map [here](https://map.anna.lgbt).


reiichi001:

Please note that all new plugins need to go to the testing repository. If/when merged, you can then move it to stable. We cannot make exceptions or other devs will also ask for exceptions, which defeats the purpose of everything starting in  testing first.

There were some concerns I had with how the hashed character data was going to be handled, based on some earlier PAC discussions before the source code was available. While I'm still of the mind that this is a bit tricky due to it being impossible for a user to ask for their data to be removed/not uploaded, as there is no was for an unknowing user to consent to having their data uploaded, the current hashing scheme that requires name, world, and customize data is suitably difficult enough to prevent someone from easily engineering a key to then hash to then try to locate a character out of game. (At least in my opinion. There's been a lot of PAC discussion on this and I expect I will not be the only one commenting on the PR).

I also tried to test the site on mobile (iOS), and it was largely unusable. The settings block covers most of the screen and cannot be moved. All interactions with it have no effect. (I cannot change a list, cannot press a button, or load the About text). The bottom view that shows player race/gender key is also cut off. 
image

It's my assumption that this is a primarily desktop-oriented site, but if mobile is supposed to be working, it may need a warning for now.

lojewalo:
> suitably difficult enough to prevent someone from easily engineering a key to then hash to then try to locate a character out of game

Keep in mind that this is only relevant if the backend database somehow gets dumped to the public. The hashed key being sent to the server is never exposed to users of the service. A separate key is generated using a different (random and hidden) salt, the hash received by the server, and the user's current territory, and that is what is exposed to users of the website.

In addition, all data that hasn't been updated in over an hour is purged from the database, as well.

> I also tried to test the site on mobile (iOS), and it was largely unusable.

Yes, I haven't looked at making it mobile-friendly yet. Is this a PAC concern or just something you wanted to report separately?

> Please note that all new plugins need to go to the testing repository.

Well I hate this, but okay, fine. Shall I just move the plugin dir to the testing dir and force-push again? This sure defeats the purpose of me having people test it before PR lmao

reiichi001:
> > I also tried to test the site on mobile (iOS), and it was largely unusable.
> 
> Yes, I haven't looked at making it mobile-friendly yet. Is this a PAC concern or just something you wanted to report separately?

Personal concern on my part. This is/was not PAC-blocking.


> > Please note that all new plugins need to go to the testing repository.
> 
> Well I hate this, but okay, fine. Shall I just move the plugin dir to the testing dir and force-push again? This sure defeats the purpose of me having people test it before PR lmao

Moving it over to the Testing folder should handle anything needed. In the previous instance of someone trying to PR to stable initially, they'd had their plugin as a custom repo option for some time. They were still required to send it to Testing first. There's no guidelines set for how long a plugin needs to be in testing, although it's still recommended to give it a few days minimum, just in case. ...as this is mostly an uploader and not displaying in-game, there's not much logic to the plugin itself and it's unlikely you'll need to make any code changes to it based on potential plugin tester feedback. (Unless the front-end site requires it)

ackwell:
this isn't a yes or no official post, but rather an update on status and opinion in general.

overall, the air remains uneasy around this idea, even with the updated hashing.
below, i'll collate some general opinions that i have seen - these are not representative of the pac as a _whole_, but may advise on where the vibe is sitting.

- client code is understandably quite simple, and has been reviewed already
	- you've mentioned you're holding back some updates - we would be interested in what they are, as it may have some impact on opinion
	- there remains unease around some of the information being sent, such as fc tag and full breadth of customisation array (unhashed)
- more than zero people were notably uncomfortable with recent usage of the plugin to "snoop" on known individual gameplay, without consent
	- it is understood that current impl does not (seem) to distinguish between instances, and as such this is somewhat offset by scale naturally obfuscating groupings, however that _relying_ on scale is an uncomfortable position
	- the lack of ability to completely opt out was noted and said to be uncomfortable
		- it is understood that this is far from the only project with snooping-like capabilities (fflogs is regularly brought up), but that has not proven to reduce discomfort
- the discrepancy between data that is uploaded and that which is sent to the map frontend is offputting
	- why upload that which is not used
	- from a zero-trust perspective, this gives a theoretical server owner more information than is let on by visible usage without looking at source
- it is understood that alternatives that provide less individual information (such as heatmaps/estimations) have been proposed and turned down
	- this isn't a negative, just an acknowledgement that something that would go well towards clearing the unease is known to be undesired

in summary; it's generally agreed that it's a cool idea that contains risk, and the balance between the coolness, the genuine usefulness (or lack thereof), and the inherent risks seems unfavourably skewed.
this is not to suggest that cool is bad - indeed, there's many throwaway/silly plugins within the system as-stands - just that the risks inherent to the current implementation don't balance out.

i do not intend to suggest alternatives / workarounds in this post, simply current status.
the specific text of this post hasn't been workshopped - it's been acknowledged that we should probably post _something_ for the current status, and i've collated this based on what i've read.

---

gods you'd think i'd learn to not write longer-form comments in text editor first, that turned into a veritable essay sorry

lojewalo:
Thanks for the update, ack.

> you've mentioned you're holding back some updates - we would be interested in what they are, as it may have some impact on opinion

Mostly minor updates like only using the gzipped body if it's actually a size savings. The other feature I added was to include the party ID from `IPartyList`, which gets salted and hashed with territory to allow the map to show individual parties within instances.

You mention that the plugin doesn't know about instances and so lots of people being in an instance provides additional obscurity, and this would make that no longer applicable. However, I find it confusing why this would matter, since it remains difficult to identify people no matter if they're being viewed as one party or one hundred parties.

> there remains unease around some of the information being sent, such as fc tag and full breadth of customisation array (unhashed)

The FC tag is useless, and I may as well remove it from the upload. I'm never going to display it, as I think it's too identifiable. I was thinking of displaying some friendly hashed version of it, but I decided it's not worth it. The customize array is not exposed to users (aside from a few points like race and gender). It remains in the schema for future uses like statistics about people in a territory, etc.

> more than zero people were notably uncomfortable with recent usage of the plugin to "snoop" on known individual gameplay, without consent 

Considering there's exactly three people I know of with it installed (one being myself), and I'm actively working on it, it's very easy for me to link together who is who (though sometimes it's not so simple). It's much less easy if you don't know exactly who has it installed and more than three people have it installed.

> the lack of ability to completely opt out was noted and said to be uncomfortable 

It's not very feasible to offer an opt out, as I'd have to know the name, home world, and full customize array of anyone wanting to opt out, and if that ever changes, they're no longer opted out. I also personally don't see the point of one, since it's very difficult to identify and follow one person anyway. Considering that the server doesn't keep historical data or old data, I don't find it a pressing issue.

> the discrepancy between data that is uploaded and that which is sent to the map frontend is offputting 

It's literally all used except for FC (and I guess rotation, but I would like to add an arrow eventually), so I'm not sure what you could possibly be referring to here.

> it is understood that alternatives that provide less individual information (such as heatmaps/estimations) have been proposed and turned down 

Yes, since they don't provide the same thing.

---

I am a very privacy-conscious person, as has been made fun of plenty in goat place, and I see absolutely no issue with this data and service. Considering the amount of work I have put into this plugin and website to make it *not* friendly to people using it for nefarious purposes and to make it suitably anonymous, I'm honestly baffled by the negative disposition. Except for the case of right now with a small user base of people I know having the plugin installed, you can't reasonably find anyone without having found them first in game. Once you do, you can't track them into different zones.

There is a distinct lack of risks listed, which I'm not faulting you for, but I'd like to know if there are actually any that people are concerned about and bringing up. I think there is an inordinate amount of pearl-clutching and very little real risk, but I'd be happy to consider any actual risks you present.

reiichi001:
> I am a very privacy-conscious person, as has been made fun of plenty in goat place, and I see absolutely no issue with this data and service. Considering the amount of work I have put into this plugin and website to make it not friendly to people using it for nefarious purposes and to make it suitably anonymous, I'm honestly baffled by the negative disposition.

As with any new plugin, there is an inherent risk that frontloaded development effort may not pan out. "I've put a lot of work into this plugin" cannot be factored as part of review. On a similar note, a plugin concept can be seen as interesting or cool or "this would be nice to have if SE did it" while also being something that we don't think could be done feasibly as a third party implementation reliant on user contributions. We do have to take into consideration how the plugin could be perceived in terms of user data and safety, how users may feel about its use, and whether it adheres to our plugin guidelines. 

> There is a distinct lack of risks listed, which I'm not faulting you for, but I'd like to know if there are actually any that people are concerned about and bringing up. I think there is an inordinate amount of pearl-clutching and very little real risk, but I'd be happy to consider any actual risks you present.

With the limited userbase, perhaps even *because* of it, in cases where there is a single party or a single uploader, anonymity starts to break down because it becomes much easier to find commonalities and consistent traits that can be used to track a player. Especially in cases where an external person is aware of party composition.

I'm personally very against the FC tag being included, and if something like party or alliance state is added, that becomes a hard no on my part as this leads to additional identifying factors.

One of the concerns expressed in our review channels is having full customize data being saved to the server when it's not being used. In nearly all cases, having a world ID, race, gender, job/level, and some other "unique" factor like clan or height percentage would be more than enough to prevent conflicts without providing your database full customize data. Multiple members are wary about this.

Having a single contributor on a given zone is also a major concern of mine. To this end, it might also be worth requiring X number of contributors and/or Y number of players found before presenting this information to the website's front-end. Like if there aren't at least 2 distinct parties in a duty zone, treat it as if there's nobody there. Or put a message like "not enough contributors to remain anonymous." 2 is just a sample number, and I think that it would need to be higher for instanced content, while it could be lower for something like a city zone that often has far more characters in it.

Having more contributors would go a long way into making some of the users feel more comfortable, as that leads to anonymity through numbers, ignoring the edge case of too few contributors is the breaking point.

I invite additional members of PAC to state their opinions and this may be a case where we need ask non-PAC people to provide their opinions too, as this is a plugin that affects ALL players, whether they use plugins or not. The non-PAC people I've discussed this with have similar concerns as I do.

lojewalo:
> With the limited userbase, perhaps even because of it

This doesn't merit response considering it's just "few users is easy to deanonymise", which is the entire point of trying to get more users.

> I'm personally very against the FC tag being included

Already said I'm going to remove it.

> and if something like party or alliance state is added, that becomes a hard no on my part as this leads to additional identifying factors.

:roll_eyes: Don't have to add it, but I really struggle to understand the harm in seeing a party in a trial versus every single party in the trial.

> One of the concerns expressed in our review channels is having full customize data being saved to the server when it's not being used. In nearly all cases, having a world ID, race, gender, job/level, and some other "unique" factor like clan or height percentage would be more than enough to prevent conflicts without providing your database full customize data. Multiple members are wary about this.

It's not used for deduplication; the plan is to use it for statistics. I am confused about why this is an issue considering that it's not exposed. Afraid I'm going to somehow use it to track people myself or something? This is some ridiculously-high scrutiny in my opinion. Did you know that PlayerTrack keeps track of every customize array you ever encounter? Is this really worse?

> Having a single contributor on a given zone is also a major concern of mine. To this end, it might also be worth requiring X number of contributors and/or Y number of players found before presenting this information to the website's front-end. Like if there aren't at least 2 distinct parties in a duty zone, treat it as if there's nobody there. Or put a message like "not enough contributors to remain anonymous." 2 is just a sample number, and I think that it would need to be higher for instanced content, while it could be lower for something like a city zone that often has far more characters in it.

What exactly does this prevent?

> Having more contributors would go a long way into making some of the users feel more comfortable

*gestures vaguely at PR*

---

I'm tired of having to justify and argue my case. Do whatever you want; it will remain as a 3pp even if you say no.

lmcintyre:
Unfortunately, your plugin submission has been denied. PAC is not comfortable with the level of detail available to users of the website. Some are uncomfortable with the ability to track others that do not consent to participate in this data collection. It is understood by the PAC that reducing collected data in efforts to appease these concerns would diminish the purpose of the plugin, and thus we cannot come to a compromise on a way to proceed. Regardless, thank you for your submission, your patience through the process, and apologies for taking your time just to arrive at this outcome.

PR #2644, opened 2023-10-08
[Testing] PlayerTrack

kalilistic:
- Fix the issue where the window would keep getting bigger on non-default global scales.
- Move backups to %appdata%\XIVLauncher\playerTrackBackups to avoid them being deleted with config data.
- Fix the issue where notes weren't saving.
- Add the ability to show warnings in the migration window.
- Attempt to migrate older or unrecognized versions of litedb instead of failing immediately.
- Warn and skip if can't clean up old backups instead of failing immediately.
- Warn and skip up to 10 players if run into errors instead of failing immediately.
- Fix back-end ci issue with toml.

### DIFF
--- a/testing/live/PlayerTrack/manifest.toml
+++ b/testing/live/PlayerTrack/manifest.toml
@@ -2,4 +2,4 @@
 repository = "https://github.com/kalilistic/PlayerTrack.git"
 owners = [ "kalilistic" ]
 project_path = "PlayerTrack.Plugin"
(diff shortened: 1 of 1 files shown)


kalilistic:
Discussed with goat and agreed to move to playerTrackBackups as an interim solution. Will move to a new backup solution once it's available.

kalilistic:
Just to be clear since I realized my last message wasn't, this already includes the modified backup and is ready to merge. Thanks!

reiichi001 (approved):

PR #2655, opened 2023-10-09
Wholist v1.4.0.0

Blooym:
## New Features
- An optional column for displaying the distance to nearby players has been added.
- All information is now updated in real time instead of taking a few seconds to reflect changes.

## Fixes
- An issue wherein logging out would break the plugin has been resolved.
- An issue wherein the list would sometimes fail to draw and spam error messages has been resolved.
- The time to draw the nearby players interface each frame has been dramatically reduced.

All of these changes will also be included in the upcoming major update for Wholist, but they have also been backported as they were easy to do so.

Thank you to [WildWolf](https://github.com/wolfcomp) for their assistance with optimizing draw times with me at 4 in the morning.

### DIFF
--- a/stable/Wholist/manifest.toml
+++ b/stable/Wholist/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Blooym/Wholist.git"
-commit = "536bba8112134cfb3da75319a372f74242243b0a"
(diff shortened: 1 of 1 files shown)


Blooym:
Note to self: don't use preview c# syntax or plogon will fail

lmcintyre (approved):

PR #2695, opened 2023-10-13
SortaKinda v1.0.2.0

MidoriKami:
This is a testing-only release.

Rebuilt filtering logic, added option for `Inclusive And`

I would like ya'll to test that things behave the same as they always have without the new option enabled, and test that the new option behaves as expected.


MidoriKami:
Diff from latest Stable -> This Testing Version
https://github.com/MidoriKami/SortaKinda/compare/9a8b61d5a69392138566e52b81097c2d04d1124b..a13d1610b42d801f1181dcb8b9c3232fcf1dd808

reiichi001 (approved):

PR #2717, opened 2023-10-16
PartyIcons v1.1.5.0

nebel:
- Update for 6.5
- Fix context menu role assignment
- Fix a bug with party chat names during chocobo races


nebel:
This is just a maintenance release for 6.5 and not an intent to fully adopt the plugin. I believe the author (and most recent maintainers) don't have time to update it recently. However I've been using these changes locally for some time and things seem to work fine.

lmcintyre (approved):

PR #2718, opened 2023-10-16
[Testing] PlayerTrack

kalilistic:
- Adjust migration to try and better handle bad data
- Add more logging to migration errors
- Fix display issues from localization changes
- Update with latest translations
- Fix backup location for Wine


kalilistic:
Probably not worth fixing retroactively, but this diff is registering as big since git thinks one of my whole files is new. I assume line-ending / whitespace antics.

image

lmcintyre (approved):

PR #2775, opened 2023-10-26
Mappy v1.0.5.2

MidoriKami:
Added Area Circle Animations, certain actions will cause a expanding and contracting circle to show to help you more easily find specific markers.

You can now zoom in and out while using the Quest List, as long as the cursor isn't hovering over the quest list.

Moderate performance improvement, (20-35% faster code)


MidoriKami:
This is a release to stable from testing, with one additional diff due to sudden unexpected ClientStructs breaking change
Actual diff: https://github.com/MidoriKami/Mappy/compare/f4558d7dc4d1b4994032023ed23614f94c552fcb..6866e15199949c9c652396242be3296ddda8be45

reiichi001 (approved):

PR #2780, opened 2023-10-26
Mappy v1.0.5.5

MidoriKami:
Fix fates not updating correctly


reiichi001 (approved):

MidoriKami:
Sorry about that, I realized moments after submitting the initial PR that a couple fates still weren't working quite right, so I had to try something else to fix it.

karashiiro (approved):

PR #2785, opened 2023-10-27
HotbarUnlimited v1.0.0.1

MidoriKami:
Adds support for hudlayouts, each layout will use and save its own config file


MidoriKami:
bleatbot, rebuild

reiichi001 (approved):

PR #2811, opened 2023-10-31
[Testing] Distance 1.1.0.0

PunishedPineapple:
This is a testing release because of a pretty large number of changes.  If you encounter any issues, or have feedback about the new feature, ***please open an issue on the plugin's GitHub repo***.  I will not see any other issue reports.

- Added a custom arcs feature to match custom widgets.  *By default, there are no custom arcs enabled.*  If you want to use this feature, go to the custom arcs tab in plugin settings, and click the '+' button to add one.  The default arc added this way is set up to show max melee range on the targeted enemy, but you can add and configure these for pretty much whatever you want (other players, yourself, etc., filtered by job, target type, and more, and colorable/fadeable based on distance).
- Brand new config UI.  The number of options was getting unwieldy in the old UI.
- Added job filters to all distance displays.  You can limit widgets, nameplates, and arcs to only show for certain jobs.  By default, these are set to all battle jobs.
- Added filters for in and out of instance.
- Added existing filter types to display types that did not have them.
- Improved nameplate node setup to more closely match the base game.  They still do not match the skinny text used on the "new"-style nameplates, but that is in the works.
- Fixed a bug that caused nameplate text nodes to not be properly cleaned up during plugin unload.
- Updated logging.
- Significant code cleanup and reorganization.


PunishedPineapple:
> Build system error: DalamudPackager output not found, make sure it is installed

I am unsure why this error occurred; I double checked the version of the package, and it is the latest version, and the same as what was used on the previous release on stable.

Is the `Move Channel` label something that I need address in some way, or is it for a reviewer?

NotNite:
Delete `DalamudPackager.targets`.

philpax (approved):

PR #2819, opened 2023-11-01
Snooper 1.2.1.1

Maia-Everett:
* Snooper will no longer play a sound when you send a tell. This was never intended. ([#18](https://github.com/Maia-Everett/dalamud-snooper/issues/18))
* If saving logs to disk is disabled, Snooper will no longer load already existing logs from disk.
* Reduced plugin log spam, fixing an issue where exception stack traces were logged for normal operation.

### DIFF
--- a/stable/Snooper/manifest.toml
+++ b/stable/Snooper/manifest.toml
@@ -1,20 +1,10 @@
 [plugin]
 repository = "https://github.com/Maia-Everett/dalamud-snooper.git"
-commit = "6fb2273f9d4fa665dc83f8e9e446fda747493f0f"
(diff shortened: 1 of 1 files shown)


Maia-Everett:
The diff looks bigger than it is because of a whitespace formatting commit. Here's the real diff:

https://github.com/Maia-Everett/dalamud-snooper/compare/f0ae06a7e3cacb47ee528df0c3e76cf7ce52b3f8..bdd2eccde81985019508604851b0229c5df4f19f

reiichi001 (approved):

PR #2859, opened 2023-11-05
NoTankYou v6.0.1.7

MidoriKami:
Automatically disable warnings for players with "Role-Playing" status.

Added sekret new feature. You'll just have to figure out what it is on your own ;)


MidoriKami:
Secret new feature is an auto-warning suppression system that disables NoTankYou warnings for other players if that player has had an active warning continuously for at least 60 seconds (by default).

There is no UI to enable this feature, users must edit the configuration file manually and reload the plugin to enable it.

reiichi001 (approved):

PR #2893, opened 2023-11-14
Tracky 1.3.4.0

Infiziert90:
- More crowd sourced data added (70k records)
- Add missing 4.0 coffer item
- Improve plugin load time
- Ensure that configs have loaded properly before working with them
- Shortend bunny timer again
  - If there is issue with missed bunny chests again, please poke me


Infiziert90:
Real diff is around 150 lines

philpax (approved):

PR #2919, opened 2023-11-20
Tracky 1.3.4.2

Infiziert90:
[Fixes]
- Prevent random stutter with Desynthesis UI open
- Eliminate crash with invalid data on Desynthesis reward search
- Don't save invalid data for Desynthesis of Collectables 

[UI]
- Redesign Desynthesis Search

Note:
Real diff is 360 lines added


philpax (approved):
Appreciate the mention of the real diff length!

PR #2921, opened 2023-11-22
Mappy v1.0.7.1

MidoriKami:
Automatically disables integrations when vanilla map is open
Added chat commands to open Mappy window
IPC Changes


MidoriKami:
About half of this diff is a demo window that lets the user preview how the IPC works

reiichi001 (approved):

PR #2924, opened 2023-11-22
Simple Tweaks 1.9.3.0

Caraxi:
***General Changes***
- Added optional metrics collection
	 - Everyone will be given a chance to opt in or our to the collection of a list of enabled tweaks when first accessing the simple tweaks config window.
	 - Anyone who preemptivly opted out will not see the notice as they already made the choice.
	 - No information will be collected until the 'Allow collection' button is pressed.

***New Tweaks***
- **`Casting Text Visibility`** - Change the font size, color, and background of the casting text. *(img)*

- **`Chat Sounds Everywhere`** - Enables <se.#> chat sounds everywhere, regardless of channel. *(Asriel)*

- **`Duty Timer`** - When completing a duty, tells you how much time the duty took. *(MidoriKami)*

- **`Emote Log Status in Server Bar`** - Show the emote log status in the server bar. *(Sythiri)*

- **`Show Glamour in Item Name`** - Displays the glamoured item name underneath the real item name. *(Enriath)*

***Tweak Changes***
- **`Track Faded Orchestrion Rolls`** - Added tracking for faded rolls with multiple crafts. *(KazWolfe)*

### DIFF
--- a/stable/SimpleTweaksPlugin/manifest.toml
+++ b/stable/SimpleTweaksPlugin/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Caraxi/SimpleTweaksPlugin.git"
-commit = "180f0a3e9fd3b6c1ed5ff686467478891a6e2175"
+commit = "ee31facee57461d7f71c97b60fea55db547d3892"
 owners = [
     "Caraxi",
 ]
--- a/stable/SimpleTweaksPlugin/manifest.toml
+++ b/stable/SimpleTweaksPlugin/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Caraxi/SimpleTweaksPlugin.git"
-commit = "ee31facee57461d7f71c97b60fea55db547d3892"
+commit = "05841ed98feb5381b4c3db63c0ff47ae28f8a15c"
 owners = [
     "Caraxi",
 ]


Caraxi:
This update includes a system for collecting information on which tweaks users have enabled. Information is mostly anonymous, using a key generated from the hash of the oldest character the player ever logged in with (Oldest `FFXIV_CHR` dir)

Information collected will be visible here: https://caraxi.grafana.net/public-dashboards/731a38e252b24e0c8de4cb68be5058ea

goaaats:
Is there a specific reason you're using character IDs, instead of persisting a random GUID? The rest seems good, thanks for the transparency.

We are asking, since not allowing devs the ability to identify characters that use their plugins in-game with metrics, if it isn't necessary for the operation of the plugin, might be something we're considering for a "rule" on telemetry. There currently isn't one and we'd like some feedback on approaches - not trying to insinuate that you are someone that would do this, of course.

philpax:
goat beat me to it (we're not sure about the use of hashed CID over a GUID), but aside from that, I've reviewed this and it's ready to go. Will merge once we've agreed on this 👍

Caraxi:
I used a hashed character id to prevent double counting someone in the event they clear their simple tweaks config, since it would generate the same key every time. Its absolutely not necessary and I hadn't considered that it would be possible to determine if a specific character id used simple tweaks by just generating the id myself (I blame doing it all after work) so I will change it to use a completely random id instead, since double counting isn't really that bad all things considered

philpax:
Awesome, thank you!

philpax (approved):

PR #2934, opened 2023-11-25
[Testing] update scout helper to v0.2.0.0

dit-zy:
this is a somewhat big update because i got tunnel vision and didn't try to release more incrementally >_< (i will do that going forward!)

* there are some substantial UI additions, and a large amount of supporting code
* added a new copy mode allowing copying of a whole template with variables
* adding zw3lf as an owner, given that we are collaborating a lot between scout helper and coord importer

### DIFF
--- a/testing/live/ScoutHelper/manifest.toml
+++ b/testing/live/ScoutHelper/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/im-scared/scout-helper.git"
-commit = "be3449ec218b4466f4b0a63a5fd6b4cc029764d2"
+commit = "4231870fe1d3deb5421f3fee9535eca4c7f3ff5d"
 owners = ["dit-zy", "zw3lf"]
 project_path = "ScoutHelper"
 changelog = "added a full-text mode for easier sharing of tracker link info"
--- a/testing/live/ScoutHelper/manifest.toml
+++ b/testing/live/ScoutHelper/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/im-scared/scout-helper.git"
-commit = "f2162d8dbadd24cd36898269e997fc28ee329539"
-owners = ["dit-zy"]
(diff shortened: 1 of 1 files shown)


dit-zy:
;-; i don't know what that error means. could it be the `https://example.com` i have as an example in the code?

NotNite:
Sorry about that, Plogon errors aren't the best - I'd suspect it's from the Git submodule you have using Git over SSH. Try switching it to HTTPS.

dit-zy:
ah how interesting! thank you, i fixed!

NotNite:
Uh, interesting! That might be a bug in submodule restoration on the build system. Likely not your fault.

dit-zy:
>_<

i did notice that ottergui has its own submodule for otterguiinternal that is using ssh instead of https. i wonder if that's relevant

philpax:
Huh. That's pretty weird - it looks like the only point of difference that I can see is that Otter has `branch = main` in his `.gitmodules` (<https://github.com/Ottermandias/GatherBuddy/blob/main/.gitmodules>). Can you try that and see if it makes any difference? (I wouldn't expect it to, but...)

dit-zy:
oh interesting. i'll try that next. otter pushed out a change to use https with his own submodule so we're seeing if that works

Ottermandias:
> Huh. That's pretty weird - it looks like the only point of difference that I can see is that Otter has `branch = main` in his `.gitmodules` (https://github.com/Ottermandias/GatherBuddy/blob/main/.gitmodules). Can you try that and see if it makes any difference? (I wouldn't expect it to, but...)

GatherBuddy does not yet use a version of OtterGui that has submodules though.

dit-zy:
bleatbot, rebuild

dit-zy:
bleatbot, rebuild

dit-zy:
closing this PR because it's a mess of changes and i don't want any reviewers to be confused. will open a new one shortly with the final, working commit hash >_<

PR #2938, opened 2023-11-26
Currency Tracker 1.3.2.0

AtmoOmen:
## UI Adjustments

- **Record Settings**: modified from a pop-up window to a stand-alone window with adjusted description text and help text for all options.
- **Export**: now located under the Record Options section.
- **Button Adjustments**: removed GitHub button and adjusted the appearance of the Data Folders, Need Help, and Languages buttons.
- **Width Offset**: Added the upper limit.
- **Amount - Change Plot**: changed from bar plot to line plot.
- **Data Tables**: adjusted default width weights for columns, upward weights for Time column and downward weights for Location column.

## Functionality Adjustments

- **Tracker**: lifted the call access between modules to each other, each module is now completely independent of each other.
- **Record Teleport Costs Module**: adjusted the logic to fix an issue where the teleportation start location was incorrectly recorded due to too fast map crossing. The module is now split into two modules, Record Teleportation Costs and Record Special Teleportation Costs.
- **Record Quest Rewards Module**: Adjusted the logic to fix the issue of not being able to properly judge quest completion due to player-specific unreasonable actions.
- **Record Exchange Results Module**: This module is now split into two modules, Record Exchange Results and Record Special Exchange Results. The logic of these two modules has been adjusted, and the problem of error reporting when the plug-in is enabled in a specific window opening state has been fixed.
- **Record Mob Drops Module**: This module has been added to allow logging of monster drops (may be inaccurate in case of multiple monsters).
- **Plots Window**: adjusted the logic of the right-click function for each chart, now using ImPlot native logic instead.
- **Add Custom Currency**: Fixed the unresponsive issue of clearing the list of items to be selected in the search box after using Hanyu Pinyin for search.
- **Currency and Currency Icons**: Adjusted the logic of getting all currencies and currency icons.
- **Select All and Inverse Selection Function**: Fixed the problem that the data table did not correctly reflect the selected status after use.

### DIFF
--- a/stable/CurrencyTracker/manifest.toml
+++ b/stable/CurrencyTracker/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/AtmoOmen/CurrencyTracker.git"
-commit = "640320e30ac7b170be23a044997d2dc39a6d6989"
+commit = "2ef7d5a21c7abba788f1b983f2ff06e17aa1831a"
 owners = ["AtmoOmen"]
 project_path = "CurrencyTracker"
 changelog = "Complete code refactoring and code cleaning"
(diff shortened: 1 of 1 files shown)


philpax:
This might take some time to review, we'll get to it asap

AtmoOmen:
Optimized the issue of long plugin loading time

philpax (approved):

PR #2945, opened 2023-11-27
[Testing] MacroMate (v1.0.6.0)

grittyfrog:
Changes:

- Add Macro search bar
- Add "PvP State" Condition


philpax:
fyi:
`State.IN_PVP_NO_WOLVES_DEN => "In PvP (exclude Wovles Den)", - presumably that's "Wolves' Den"?

grittyfrog:
:facepalm: Thanks for the FYI, will fix shortly.

PR #2951, opened 2023-11-28
[Testing] AutoTimer (v1.0.0.1)

s5bug:
Adds hint text to the bar background selector in the configuration window clarifying how it should be interpreted on Ninja, as well as fixes a bug where the base GCD of a class was uncalculatable due to searching for `"Spell"`/`"Weaponskill"` in all game languages instead of using the locale-independent spreadsheet row ID.


s5bug:
Hmm, that's not right... I must have to rebase my branch. Give me a moment...

PR #2964, opened 2023-12-02
HotbarUnlimited v1.0.2.2

MidoriKami:
nofranz

Disable in PvP

I originally wanted this to be disabled in PvP but at the time the code to disable in PvP didn't quite work right.
It now disables properly in PvP as intended.

### DIFF
--- a/stable/HotbarUnlimited/manifest.toml
+++ b/stable/HotbarUnlimited/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/HotbarUnlimited.git"
-commit = "5a0df592ffc5a7554537ee0f69af231fc5b7a463"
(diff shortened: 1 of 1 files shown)


lmcintyre:
Hey @MidoriKami, on PAC's side, I just want you to be aware that we don't think that this plugin needs to be disabled in PvP. If you are doing this out of caution for PAC, I wanted to let you know that we don't feel it's strictly necessary.

Let me know if you still want this merged and I'll get it done.

MidoriKami:
Yes please, the only reason it wasn't disabled from the start was technical issues, I've since improved the systems and can achieve what I originally intended.

The only two plugins of mine I want enabled in pvp is chillframes and sortakinda

lmcintyre (approved):

PR #2978, opened 2023-12-06
ChillFrames v2.1.1.1

MidoriKami:
Fix using settings that are no longer available.


MidoriKami:
The diff was too small, had to fix it.

PR #2997, opened 2023-12-12
SortaKinda v1.0.3.1

MidoriKami:
Add `Repairable` property filter


MidoriKami:
A majority of this diff will be from changing braces from
```cs
public void Function()
{

}
```
to 
```cs
public void Function() {

}
```

I only applied this change to the files I needed to modify for this feature.

philpax (approved):

PR #3021, opened 2023-12-16
Collections v1.0.1.4

Seventhxiv:
NO functional changes!

This is just a bunch of Refactoring I wanted to do for a while (removing unused files, renaming stuff, moving files around, cleaning some interfaces)


philpax (approved):
Thanks for splitting these out from functional changes!

PR #3047, opened 2023-12-26
Simple Tweaks 1.9.5.0

Caraxi:
***General Changes***
- The Simple Tweaks config window will now be decorated during various festive periods.
	 - This can be disabled in the 'General Options' tab.
	 - You can also set a permanent decoration to be used all year round.

***New Tweaks***
- **`Remove Client Limit`** - Allows opening more than the standard two FFXIV clients.

- **`Reposition Cutscene Dialogue Box`** - Allows setting a custom position for dialogue boxes used in cutscenes.

### DIFF
--- a/stable/SimpleTweaksPlugin/manifest.toml
+++ b/stable/SimpleTweaksPlugin/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Caraxi/SimpleTweaksPlugin.git"
-commit = "ec747689a0c230fab8550f75a71e392a7210936f"
+commit = "84003865465f1ee516f7fc78c2840753660e6077"
 owners = [
     "Caraxi",
 ]


Caraxi:
Regarding the "Pending Rules Compliance" which I assume relates to the `Remove Client Limit` tweak:

This tweak does not allow multiple logins with the same service account, it purely allows users to access the additional accounts they have literally paid to have access to. Nothing stops people from running the game in a sandbox to achieve the same result, this only decreases the amount of effort people need to put in to **access the accounts they pay $20 a month to have access to**. If Square Enix weren't so stupid they would never have added this limit in the first place.

philpax:
Yep, it's regarding that - we agree in spirit with your argument, but we're trying to confirm that we're all on the same page with regards to the interpretation of our own rules and SE's intent here. This is taking a bit longer than usual because a few people are out on holiday - sorry about the delay, we'll try to get this sorted ASAP!

philpax (approved):
All sorted. We'd appreciate it if you could add a note about the complications of a multi-client setup (plugins breaking, limited support from us, etc) in the future, but we're not going to hold up the submission on that. Thanks!

PR #3054, opened 2023-12-28
[Testing] Update Visibility (testing/live)

SheepGoMeh:
Implemented a proximity setting which lets you hide entities based on their proximity to you and the threshold (how many entities are within the proximity radius)

Please report any issues via feedback/Discord/GitHub Issues


SheepGoMeh:
Wrong net sdk.

PR #3056, opened 2023-12-28
Currency Tracker 1.3.3.4

AtmoOmen:
- All UI elements now can adapt to scaling.
- Now you can update the localization files in-plugin from the cloud.
- New feature: Interval Alert - You can set unlimited customizable intervals for currencies' Change/Amount to inform you.

Real diffs: 628 lines (86 lines from the two ``.resx`` files and 792 lines for only indentation adjustment from two files)


AtmoOmen:
why this build failed and it seems that nothing wrong with my code...

reiichi001 (approved):

PR #3064, opened 2023-12-31
SonarPlugin v0.6.1.0

rreminy:
nofranz

`SonarResources` has also been open sourced here. This project generate all resources for Sonar.


NotNite:
This looks like a bug in Plogon lockfile code that Bluefissure also ran into with OBSPlugin, where lockfile .NET versions are stricter than they need to be. For now I'd just delete any lockfile that contains net8.0 and regenerate them.

rreminy:
Yes, I'm in progress but I'm instead removing net8.0 from TargetFrameworks

rreminy:
Artifact checked, this is ready now.

reiichi001 (approved):

PR #3069, opened 2024-01-02
The Great Controller Hud Switcher 1.0.0.0

CallumCarmicheal:
A simple plugin that automatically switches between controller / mouse mode with a hud layout switch when ever WASD is pressed or the controller joysticks are moved.

To change hud layouts the command `/hudlayout <no>` is used, there is a built in limiter to ensure this only gets called once per second and the keyboard/controller is polled every 250ms to reduce cpu usage to a minimum.


NotNite:
The folder name in the repository must match the plugin InternalName.

CallumCarmicheal:
I see thanks :)

reiichi001:

In the meantime, please move this over to the Testing folder in D17. All new plugins have to start there first. If/when approved, you can then move it to stable later.

MidoriKami:
Is the joystick sensitivity adjustable? What if someone is playing with a controller that has an unexpected amount of drift?

reiichi001:
> Is the joystick sensitivity adjustable? What if someone is playing with a controller that has an unexpected amount of drift?

Wouldn't/shouldn't that be the same as FFXIV's own controller calibration? ...because if a joystick has enough drift to spin the camera or make a character move, that's far more impacting than the HUD changing to controller mode.

CallumCarmicheal:
> Is the joystick sensitivity adjustable? What if someone is playing with a controller that has an unexpected amount of drift?

As reiichi001 said, the deadzone configuration is controlled by FFXIV which is labelled as "Controller Sensitivity". That dictates the deadzones for IGamepadState, additionally there is another configuration letting the user set at what value to swap to controller mode at.

Here is a grab of the configuration screen:
image

goaaats:
bleatbot, rebuild

philpax:
Hi, apologies for the delay on this. I'm assuming responsibility for this PR and will work with the rest of the team to get this in ASAP.

philpax:
Thanks to @Kurochi51 for testing this! This is what they had to say:

> registers the main command twice and removes it twice
> everything else seems fine, timeout included, aside of the Layout options being cut off https://i.imgur.com/WMwUZPD.png
> this is at 1440p with font scale 14.5pt

We'll merge it in now.

philpax (approved):

PR #3072, opened 2024-01-03
[Testing] HUDManager 2.5.18.3

nebel:
- Fixes an issue where the "Style" option for the Status Effect element didn't behave correctly in some cases. The "Style" option has been split into "Grouping", which determines if status effects are displayed in a single element or split into multiple groups, and "Alignment", which affects the justification setting of the single element mode. This more closely reflects how the options are stored internally by the game and should be more reliable.
- Fixes an issue where changing the layout of a split status effect group wasn't reflected properly in the UI preview.


nebel:
bleatbot, rebuild

reiichi001 (approved):

PR #3074, opened 2024-01-04
Update Eorzea Votes to 1.2.0

lojewalo:
- Wrap text for inactive questions
- Add pagination for inactive questions


lojewalo:
Daily reminder that half the code in this diff is Rust, so I'd call this small not mid.

reiichi001 (approved):

PR #3092, opened 2024-01-09
Who's Talking 0.6.7.0

sersorrel:
promote to stable, since nobody complained recently

I should've done this ages ago, I was waiting to get round to a UI revamp but I would honestly be surprised if I finished that before dawntrail at this point

also a minor addition to the plugin description and adding a missing TreePop

(what's best practice for handling parallel testing/stable tracks? do I need to bump both channels, or can I leave testing alone and people will get updates with a higher version number from stable automatically?)


lmcintyre:
You can leave Testing the way it is or entirely remove it. The higher version from Stable will override it. Up to you.

reiichi001 (approved):

PR #3097, opened 2024-01-10
DeepDungeonDex 2.9.2

wolfcomp:
### Bug Fixes

* change content type to local flag enum (5f2e180)
* data loading to only load local once if first attempt fails (d1e2e82)
* finalize unknown content display (b1eebe2)


wolfcomp:
found an error will fix and resubmit

PR #3111, opened 2024-01-16
EngageTimer v2.3.1.0

xorus:
- You can set alarms to play a game sound effect, change stopwatch color or display text at specified combat durations
  - Big code rewrite and a bit of optimization
  - Fix save errors when spinning color sliders like a maniac in configuration
  - Reorganized configuration file to preserve my sanity
  - "Hide original addon" now uses AddonLifecycle events instead of searching for the original countdown every frame
  - Reduce CountdownHook CPU usage by fixing a stupid event spam mistake
  - Optimize countdown display code
  - Fix the "first-draw" workaround that draws the countdown window once on plugin activation to prevent a freeze caused
    by ImGUI initializing the window does not occur when starting a countdown

### DIFF
--- a/stable/EngageTimer/manifest.toml
+++ b/stable/EngageTimer/manifest.toml
@@ -1,8 +1,14 @@
 [plugin]
 repository = "https://github.com/xorus/EngageTimer.git"
-commit = "69ff776388847692aefa0309300d3535792edda8"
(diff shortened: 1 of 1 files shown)


xorus:
Sorry for the huge diff I rewrote/reorganized a bunch of stuff. I am also so disconnected from the game at the moment I completly forgot today was patch day so I can not test this on the newest xiv patch yet, oops.

PR #3149, opened 2024-01-31
[testing/live] Fast Job Switcher 1.0.0.3

meoiswa:
Enables the player to switch jobs using their three letter acronyms as commands.


MidoriKami:
Seems this might be better suited as a tweak so you don't have to wait 1-2 months for merge

meoiswa:
I also agree it would be better suited as a Tweak, however, it was rejected by its maintainer.

MidoriKami:
Ahh I see now the note that was left on the proposed tweak.

An alternative approach then would be to expand the Command Alias tweak with the ability to import/export alias presets.
Might also be able to add presets that the user can enable by name.

For example, as seen in this **amazing** mockup
image

reiichi001:
The code is short/simple enough that I can sign off on it. I'll add this to our review process.

goaaats:
bleatbot, rebuild

goaaats:
bleatbot, rebuild

goaaats:
bleatbot, rebuild

philpax:
thx 4 new plugin

@reiichi001 on it

ok bye

philpax (approved):

PR #3184, opened 2024-02-12
[Testing] ScoutHelper v1.0.0.0

dit-zy:
this change is kind of big >_<. a bunch of changes are to the unit tests `ScoutHelperTests`. but this change adds support for a whole new scout tracker website, sirenhunts.com, so there is quite a bit added to support that. notably, though, siren doesn't need any web calls -- it's all string manipulation to generate the links with appropriate url parameters.


dit-zy:
bleatbot, rebuild

philpax (approved):

PR #3214, opened 2024-02-20
Pet Scale 1.0.0.0

Kurochi51:
nofranz

The plugin lets you make an entry for yourself, other players, and specific named players that are around you provided by a list either on request or config window opening. You can either choose `AllPets` or a specific one, then any of the 3 sizes. Once the entry is made, the plugin will set a summon that matches the specifics to the desired size.

Right now that's pretty reliant on data order, as it can be seen inside `InitData()` and everytime `Save()` gets called in `ConfigWindow`, and assuming nothing went wrong, all entries are sorted with `Other Players` being at the top, then all other names are alphabetically sorted, and then these entries are sorted again to make sure that any `AllPets` entry is at the top.

The reason for this is the order in which they're applied in-game, which is:
* General Pet - General Character
* Specific Pet - General Character
* General Pet - Specific Character
* Specific Pet - Specific Character

This bit actually gave me a bit of a headache, but this is the most concise way that I managed to come up with to avoid iterating over the data a whole bunch of times. Once a size is set, a dictionary that tracks all currently active pets and characters will be changed to prevent that specific entry to be set again. The dictionary is then cleared after a fixed period of time, in this case `TimeSpan.FromMilliseconds(500).TotalMilliseconds` and is re-populated from `CharacterManager.Instance()->BattleCharaListSpan`.

Most likely before this gets looked over and approved, I'll also add a button in the `ConfigWindow` to clear the entire dictionary, and I welcome any other suggestion.


testing/live/PetScale/manifest.toml
ackwell, line 1: nit; https://github.com/Kurochi51/PetScale/blob/main/PetScale/PetScale.cs#L156

```cs
if (!Enum.TryParse(typeof(PetModel), name, out var model) || model is not PetModel modelValue)
```

this seems like a very roundabout way to build the mapping - why not store a direct dictionary of pet id to ModelCharaId?

ackwell, line 1: nit; https://github.com/Kurochi51/PetScale/blob/main/PetScale/Windows/ConfigWindow.cs#L391

```cs
private void Save()
```

this seems duplicated with init logic. possibly worth looking into refactoring that.

ackwell (approved):
nothing blocking. few code nits.

---

i do wonder if it's worth supporting other jobs. i don't really see strong reason to limit it to just pets with existing scale handling.

---

i think there's def. some potential work to be done w/r/t the mappings that are currently defined as enums - much of those would be better suited to dictionaries or similar

testing/live/PetScale/manifest.toml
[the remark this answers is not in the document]
Kurochi51: This specific `Queue` is used to store the players that are on-screen from `BattleCharaListSpan`. For what I'm using it as, it's functionally identical to a normal `List`, both of which are backed by a dynamic array. The default capacity is purely an optimization, so the queue won't have to change it's size constantly.

Kurochi51: Actually it's the other way around, wrote this first then copy-pasted it into the init function because I had an existing config with unordered elements. If anything I can remove the init, right now it's mostly a sanity check.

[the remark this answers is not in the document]
Kurochi51: I needed to make sure the specific RowId is part of the enum, hence casting wouldn't work. I'll check out `IsDefined` tho.

Kurochi51:
I didn't limit this to any job, an it technically supports all pets right now, aka any carbuncle / egi / fairy, or chocobo. I kinda wanted the enums themselves to be the ones that control what goes in, hence I made and used them to gather the info from `Pet` exd. The only issue is that I'd have to assign some random values for sizes, since right now everything but summoner pets is already scaled at 100%

testing/live/PetScale/manifest.toml
Kurochi51: I needed 2 dictionaries that both have the name of the pet, and then a corresponding `PetModel` and the sizes.
I'm using the the dictionary with name and `PetModel` purely for the `Combo` in the `ConfigWindow`, to display the name of the pets, but save the `PetModel` in the config file. 
Where as I needed a similar dictionary with their names and their sizes grabbed from the exd to match the name of the pet to the corresponding `PetSize` enum. Thinking back on it, I could probably swap the string to a `PetModel` type, but I'm not sure it would make a difference.

philpax:
Testing completed thanks to @MKhayle!

philpax (approved):

PR #3217, opened 2024-02-21
Tick Tracker 1.4.0.0

Kurochi51:
nofranz

I've been sitting on this for a few weeks, I didn't notice anything out of order, but it's kinda difficult to test these things when you know how they're supposed to work. 

The new native ui part attaches to the frame of the hp and mp/gp bar, and will grow alongside the length of said bar. The color isn't uniform, so some parts will look darker, this is because of the underlying texture used for the image node. 

There's also some penumbra ipc to disable the native settings if `Material UI` is detected, that's because the textures that MUI comes with lack the specific part I need, and I rely on client-loaded textures.


Kurochi51:
Note: at least a third of the loc diff should be due to `.editorconfig`, `.gitignore`, and `Plugin.cs` to `TickTracker.cs`

PR #3228, opened 2024-02-24
Tilted 3.4.1.2

meoiswa:
Version 3.4.1.2
 - Added toggle to disable Tilted when GPose is in use
 - Added trigger to detect flying in a mount (separate from simply riding a mount)
 - Multiple fixes to minor math errors


goaaats:
bleatbot, rebuild

goaaats (approved):

PR #3253, opened 2024-03-02
[Testing] Ntfynder 1.1.2.0

Arcadia64:
This plugin sends a push notification using `ntfy` when a change occurs in the party (duty pop, player leaves/joins, etc.). This is intended to help people sitting in party finder for a long time doing savage/ultimate and want to go afk.

**ntfy**
https://docs.ntfy.sh/


philpax:
Hi there! This is pretty straightforward and we'll get it through the other stages of the process, but I was wondering what the reasoning was behind forking PushyFinder instead of implementing a ntfy backend for it? 

We're not opposed to having multiple plugins that do the same thing, but given that the primary changes in your fork are [the delivery mechanism and name](https://github.com/lostkagamine/PushyFinder/compare/master...Arcadia64:Ntfynder:master), it seems like users would be better served by having a single plugin that can use either mechanism.

As an aside, as you have forked PushyFinder, the MIT license requires you to retain the original license notice in addition to your new one:
> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

Thanks in advance!

Arcadia64:
Thanks for the feedback. I was considering simply adding support for ntfy to PushyFinder, but I also wanted to keep things as simple as possible for users to configure. Currently PushyFinder is suggesting a paid service (Pushover) and I think if a user sees that as the first suggestion it may scare them off. In my opinion, ntfy is the easiest and free option and should be the default. My main goal was to have the setup as simple as possible. The PushyFinder creator also mentioned they were going to add support for more push notification providers, but I don't know when that is planned for or what they intend on including.

Additionally, on the original developer's profile page they state the following.
> Unless stated otherwise, any and all public repositories on this account are unsupported. Please do not ask about how to use them, as they are personal tools, more often than not. You may still report issues, and I may get to them, but it is not a guarantee. I battle with burnout quite often, and I would like to keep the threat of falling into it to a minimum.

> That said, please respect the licenses. Most of my work is licensed MIT.

This lead me to believe even if I did add support for ntfy, it may be quite a while before it ever makes it into the plugin. And yes I will definitely fix the license. Always trying to give credit where credit is due :)

philpax:
Thanks for the quick reply! That makes sense. Have you been able to speak to @lostkagamine about this? We're happy to go ahead with this if you've made an attempt and you haven't gotten a reply.

We have an [adoption policy](https://dalamud.dev/faq/adoption/) that allows for someone to take over development on a plugin if the original author is unavailable, but my understanding is that Sylvie is still around, just busy with life in general. 

The ideal outcome for us would be a single plugin, supporting either backend (and probably defaulting to Ntfy for the reasons you've mentioned), with the ability for either one of you to maintain it. We can work something out if that's not possible for whatever reason 🙂

philpax:
Hi there!

A week has passed without an update, so I'll assume that the original author is unavailable. I'll go ahead with getting this tested so that we can get it merged in 🙂

philpax:
Tested by @NotNite, all good to go. 

philpax (approved):

PR #3256, opened 2024-03-03
Pet Nicknames v1.4.8.3

Glyceri:
[1.4.8.3]
'Give Nickname' has been renamed to '[P] Rename' (Get used to it)
The plugin icon is changed because in reality... I cannot see green properly.


philpax (approved):
not being able to see green properly would explain the previous icon, yes

PR #3272, opened 2024-03-09
[Testing] EasyTranslate v1.0.0.0

Olaren15:
This new plugin allows players to search for translations of in-game elements. It uses xivapi to search elements by name and show the users their English, French, German and Japanese names.

This plugin is really useful when playing with people who have their client in a different language!


philpax:
bleatbot, approve

philpax:
Hi there! Sorry for the delay on approving this plugin for review.

It looks like there's [a few issues with the build](https://github.com/goatcorp/DalamudPluginsD17/actions/runs/8223556700/job/22486392376); it looks like you're using C# 12, which we don't currently support. We'll likely add support for that in the future, but it might be easier to use an older version of C# in the meantime.

Olaren15:
Thanks! I'll look at it and try to fix the build errors

philpax:
Hmm, looking into why that failed, will get back to you...

NotNite:
Our build system has a slight flaw that the folder name that contains the manifest in this repository needs to match the internal name of the plugin (EasyTranslate.DalamudPlugin).

Olaren15:
Thanks a lot for the help! The build is working now :)

philpax:
Thank you for finding three bugs in our build system 😅

I've completed the code review and everything looks good, but is it possible to use Lumina instead of XIVAPI for the queries you need to do? Lumina uses the local game files, which should allow it to be lower latency / more reliable.

Olaren15:
I've looked a bit into using lumina to get the translations info, but the advantage of using xivapi is that is searches through almost all of ffxiv's content (items, quests, fates, mobs, achievements, etc...).

From what I understand, if I were to use Lumina, I would have to implement a search function for every Excel sheet. I also presume that I would have to iterate over *every* item in a sheet in order to search by name. Please correct me if i'm wrong.

So yeah I think that using xivapi and its processed/indexed data saves me a lot of complexity 😅

philpax:
Fair enough - yes, I believe you're correct. Happy to leave it as-is then :)

Will continue with the review process, then!

philpax:
All sorted! (tested by @NotNite)

We'd still like to help you switch over to Lumina at some point (including potentially building/shipping a local index), but it's not a blocking concern right now. Just reach out on Discord if you'd like to brainstorm ideas 🙂 

---

philpax (approved):

Olaren15:
Thanks a lot for approving and testing my plugin!

I think that moving to Lumina eventually would be a good move and I would appreciate help to make it happen! (I'm still a c# and Dalamud plugins noob lol)
I originally had plans to also make a website that offered the same functionality, but outside of the game (hence the over-separation of the code ^^') but I found out that something like this already existed haha (https://xorus.github.io/xivtrans/)

I would also like to join the dev channel on discord :) my discord username is olaren

PR #3286, opened 2024-03-13
Orchestrion 2.2.0.2 -> 2.2.0.3

lmcintyre:
- Fixes chat channel selection
- Fixes various localizations not being included in the plugin
- Updates the included song lists and fixes that the Chinese list was not actually included in the past
- Fixes Chinese song titles not being displayed properly in tooltips (Note that if you would like to use the plugin in Chinese on Global, you will need to set a Dalamud Font that supports Chinese, as Axis does not.)


philpax (approved):
I trust you haven't put something naughty in the data because the code looks fine

PR #3294, opened 2024-03-14
[Testing] Paggles 1.0.0.1

InvoxiPlayGames:
Changes the time in the Server Bar to use the system locale for AM/PM designation.

image
image


philpax:
bleatbot, approve

philpax:
Hi there, thanks for the submission!

This is [a relatively small change](https://github.com/InvoxiPlayGames/paggles-xiv/blob/566668ceb10977bbd71eaed8ceb67ccd462c4910/Paggles/Plugin.cs#L70-L85) to the game, which puts it in the class of things that would typically be implemented in @Caraxi's Simple Tweaks.

Is that something that the two of you would be interested in, or would you prefer for it to be maintained as a separate plugin? (Cara has limited bandwidth for maintenance, so it's up to them if this is something they can accept.)

We're okay with leaving this as a separate plugin, but ST might make for a more natural home for it. Up to y'all.

InvoxiPlayGames:
I'd prefer to maintain this as a separate plugin.

philpax:
Sure. We'll proceed as normal, then 🙂

Kurochi51:
Notes from testing:
- I noticed that on my end both ET and LT times only say `aggles`, while ST shows `paggles` just fine
- This is more of a nit than anything, but the plugin doesn't do anything if the tweak `Custom Time Formats` is enabled, an info mark or something in the settings window that mentions that would be nice.

philpax:
Thanks to Kurochi for testing! Please consider addressing their feedback when possible. In the meantime, this is good to go 🙂 

---

philpax (approved):

InvoxiPlayGames:
> I noticed that on my end both ET and LT times only say aggles, while ST shows paggles just fine

was the current Eorzea time and your local time in AM? that'd be intentional

> the plugin doesn't do anything if the tweak Custom Time Formats is enabled, an info mark or something in the settings window that mentions that would be nice.

thanks for the recommendation (and testing) - i should probably also put an info mark in the settings window if the Windows locale settings would make the plugin appear broken in locale mode for the person using it, like being set to either "a.m."/"p.m." (the game's default) or nothing at all

---

> please let us know your Discord handle and we'll grant you the role.

my Discord handle is the same as my GitHub username

Kurochi51:
Oh yeah, it was AM time, guess that makes sense lol

PR #3302, opened 2024-03-16
NeatNoter 2.5.1.0

kalilistic:
When Dawntrail releases, NeatNoter will no longer function nor receive updates. It has been on life-support since the last expansion, and unfortunately, I've been unable to find developers interested in continuing the project.

Please consider exploring alternative plugins like NOTED or other note-taking tools that might suit your needs.

For a smoother transition, I've added an export to CSV feature. Also, if you're comfortable with SQL, you have the option to access your notes database directly via LiteDB Studio. You'll find all your data in the plugin config folder.

Thank you for using NeatNoter.


philpax (approved):
😢

PR #3319, opened 2024-03-20
VFXEditor v1.8.3.0

0ceal0t:
- net8 + C#12 updates
- `.uld` editor tweaks
- Added more icons in select dialogs
- Added support for material animations
- Updated to some `.tmb` entries
- Fixed some `.mdl` parsing


philpax (approved):
I would tell you to make smaller PRs, but I know that's physically impossible

PR #3332, opened 2024-03-20
[Testing] Who's Talking 0.6.9.0

sersorrel:
nice


philpax:
Sorry, can you explain what those hardcoded credentials are?

reiichi001 (approved):

sersorrel:
> Sorry, can you explain what those hardcoded credentials are?

They're the (fake) user details returned by arRPC on RPC startup: https://github.com/OpenAsar/arrpc/blob/b4796fffe3bf1b1361cc4781024349f7a4f9400e/src/server.js#L32-L55

which I'm using to detect when I'm talking to an RPC server that doesn't actually work. (The official client doesn't appear to return user data in `READY` at all, despite what the documentation says.)

PR #3334, opened 2024-03-21
MacroMate 1.0.8.1

grittyfrog:
Changes:

- Fix issue when linking to overlapping Shared/Individual slots
- Allow setting the Link Placeholder Icon
- net8 upgrade
- use new font api


grittyfrog:
Updated changelog, turns out old import/export works fine

philpax (approved):

PR #3353, opened 2024-03-24
[Testing] Interjection 0.0.0.0

sourpuh:
- Show which casts are interruptible in the Enemy List
- Color enmity gems when enemies are targeting other tanks


lmcintyre:
@sourpuh,

Hello! Thank you for your submission and apologies about the delay in plugin approval. The PAC has concerns with this plugin related to this particular rule in the plugin guidelines:
> your plugin does not augment, alter, or interfere with combat, unless it only provides information about your own party or alliance members that is otherwise available, but represents said information differently. 

For example, your plugin offers information about other mob casts that is not readily available without targeting that mob.
In addition, your plugin offers information about each mob's aggro target and their job (is the mob targeting a player? is that player a tank?) that is not readily available without targeting that mob.

I am unsure of how you may want to rectify this if you are still interested in submitting your plugin to the main repository, but PAC is available for rules discussion in this thread should you intend to modify your plugin to adhere to these guidelines.

sourpuh:
Thanks, @lmcintyre 

There is no real way to rectify this; the plugin violates that rule by design. There are so many 1PP that violate that rule that I wasn't aware it is applied so strictly.

> Note that there are plugins on the repository that do not abide by this rule, but they have been grandfathered and similar plugins will not be allowed.

It would be nice if you documented these plugins to help developers be aware of the selective enforcement.

Thanks again!

PR #3374, opened 2024-04-01
GoodFriend v3.3.1.0

Blooym:
- Fixes delayed reconnect logic
- Attempts to fix console warning spam about accessing the object table from another thread.


karashiiro (approved):

karashiiro:
`RunOnFrameworkThread`?

PR #3399, opened 2024-04-07
[Testing] MacroMate 1.0.9.0

grittyfrog:
- Auto-translate support (copy/paste only)

This is a fairly significant change under the hood so please report any bugs while this is under testing. 

(My plan is to leave this version in testing for a week or so and then publish it to stable, since under the hood we moved from `string` to `SeString` and I want to see it work cleanly on other machines)

### DIFF
--- a/testing/live/MacroMate/manifest.toml
+++ b/testing/live/MacroMate/manifest.toml
@@ -1,6 +1,11 @@
 [plugin]
 repository = "https://github.com/grittyfrog/MacroMate.git"
-commit = "8540d3b14a2243131cbaead16431bb185e2fa44d"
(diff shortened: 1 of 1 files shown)


grittyfrog:
Not sure if the diff is accurate since I haven't been updating this manifest against stable. I believe the actual diff is `1.0.8.1` -> `1.0.9.0` so I believe this is an accurate [Diff](https://github.com/grittyfrog/MacroMate/compare/a046ed5a0786435c72287068f258bd4f2f4821a0..bed2fd1378983a3796c5b89057f542a66a4ba924).

reiichi001 (approved):

PR #3409, opened 2024-04-12
[Testing] XIVInstantMessenger 1.4.0.0

Limiana:
- Fixed an issue where first message wouldn't be saved to logs
- Fixed "Skip link opening confirmation" not working
- Added an option to suppress generic channel unread marking
- Changed font system to use Dalamud's new font selection and loading. This will cause your configured custom font to be reset and you will have to reconfigure it. 
- Added an option to split message logging per character
- Added an option to hide character's home world name from tabs
- Removed empty message and target notification
- Unless in high demand, translation feature is scheduled to be removed from the plugin


Limiana:
Probably need 8.0.4 here https://github.com/goatcorp/Plogon/blob/master/Plogon/BuildProcessor.cs#L90

Limiana:
I'm not exactly sure why it still tries to use windows targeting, I configured solution to use Ecommons configuration without forms or windows targeting

reiichi001 (approved):

PR #3434, opened 2024-04-21
ChatTwo 1.22.0

Infiziert90:
**New database backend**
- From LiteDB to SQLite
- Positive:
    - Performance improvements for both read and write times
    - First login, with enabled history, is massively speedup
    - Better error handling and hopefully less corruption issues
    - Smaller disk size
- Negative:
    - MessagePack encoded chunks, so user won't be able to manually read things in a DB viewer
- Planned:
    - Ingame database viewer with date selection and other fancy things
    - Export to CSV or TXT files
- Old LiteDB Database:
  - Can be migrated through the notification or config
  - Can be deleted in the configs database section
  - All support functions will be removed 2 months after this update dropped

**New behaviour for switching tabs**
- Tabs with input channel
    -  Switch to the selected channel

- Tabs without input channel
    - First time switched to this channel (current launch), they will stay at whatever channel is currently active
    - Afterward they will remember the previous channel and restore it on switch

**Keybind fix**
- "Ready Chat Log" and "Ready Text Command":
  - Are ignored if any vanilla text input is active
  - Are ignored if the direct chat option is selected

### DIFF
--- a/testing/live/ChatTwo/manifest.toml
+++ b/testing/live/ChatTwo/manifest.toml
@@ -5,3 +5,4 @@ owners = [
     "Infiziert90",
     "lojewalo"
 ]
+project_path = "ChatTwo"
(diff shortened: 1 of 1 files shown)


Infiziert90:
Diff from stable: https://github.com/Infiziert90/ChatTwo/compare/05fcdbe21ae447d329576df626a86a816ef8fc00..b4cb73dd0aa0f33d22ad5ede6bc8af7b47793b19

reiichi001 (approved):

PR #3440, opened 2024-04-23
Waitingway 2.0.1

WorkingRobot:
Huge rewrite! Hi, I'm Asriel, and I'm taking over this project! Waitingway now works after 2 years of laying dormant.
- ETA times have been added (locally calculated right now until I can gather more data)
- For Discord notifications, you will need to reauthenticate
- Send queue notifications to multiple Discord accounts (if that's your thing)
- All previous plugin data is now unused and will be deleted


philpax:
Hi! Just letting you know that this will be treated as a new plugin in terms of testing/code review because, well, you rewrote it 😅 

It might take a while, so you'll just have to bear with us.

philpax (approved):

PR #3458, opened 2024-04-28
[SimpleTodo] New Version 0.0.0.2

Lunalicious:
Heyu o/
As the changelog suggests, this new version fixes a funky bug, which caused the user to not be able to check or uncheck Todos from wherever in the list, but only from top to bottom... The cause was me not using ImGui IDs! Lesson learned

On top of that, I renamed the Plugin to "Simple Todo" which felt more natural for non Devs to me than "SimpleTodo".

And I refactored "some" parts for a more organized Code structure (sorry for the reviewers)
Also... please don't asked what happened with the commit history. It's a mess now :)

Thanks for your amazing work <3

### DIFF
--- a/testing/live/SimpleTodo/manifest.toml
+++ b/testing/live/SimpleTodo/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://gitlab.com/moonfox-projects/simpletodo.git"
-commit = "f4f95b0c31cd090d9c1d9745ce32eaac6cef0cb8"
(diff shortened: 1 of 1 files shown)


goaaats:
Heya, we can't merge this if you don't increase the version number of your plugin. I see that you increased it manually in the json, but the .csproj version always overrides that one, and it is still set to 0.0.0.1.

Lunalicious:
Aaah I see! Thanks for pointing that out!

philpax (approved):

PR #3462, opened 2024-04-28
[Testing] VoidBox v0.1.0.6 (init)

0x0ade:
Babby's first (technically second, but we won't talk about my reflection crimes) real attempt at getting a plogon into the official repo.

# VoidBox
### Basic plugin to add a skybox to the housing void, making jump puzzles pretty.
Based on [Ktisis](https://github.com/ktisis-tools/Ktisis), thus using the same license despite me usually not being a fan of the GPLv3.

(Strictly speaking it's just the env copy hook / signature and the accompanying structs which are copied, but credit goes to chirp for implementing it first!)

This plugin was hotglued together in my spare time to do one job and one job only. I didn't even bother matching the code style or adjusting the namespaces, and most of the time I'm touching it, it's around midnight and I want to sleep.

I've been testing this extensively with some of my friends for the past few weeks and already fixed some problems we've spotted - most notably, false positives due to how I'm checking if the skybox is the housing void that needs to be replaced. Surely there's a better way, but we haven't found any new false positives since my last update, and I'm eepy 🛌


philpax:
Hi there!

Apologies for the delay on a response; we were conferring amongst ourselves and it took some time to come to a decision.

Unfortunately, we've come to the conclusion that we have to reject this plugin. We appreciate the effort you've put into it, but VoidBox's use case is in service of out-of-spec gameplay, and having it in the official repository would serve to encourage such gameplay. We recognise that this is not in our official list of rules, and we apologise for that; it's an oversight that we'll be correcting.

We're very sorry about this, and hope you aren't discouraged!

0x0ade:
No problem, that's fine - that's what third party plugin repos exist for :sweat_smile: Thank you for the info, it's really appreciated :)

PR #3472, opened 2024-05-01
[Testing] Pet Nicknames 1.4.8.7

Glyceri:
Upgraded to the new Context Menu system.


philpax (approved):
that download code is a little odd but I trust you've thought it through

goaaats:
FYI, the texture cache thing is by design. The preferred solution to your problem would be to append an UUID to the path.

PR #3501, opened 2024-05-11
[SimpleTodo] Update to 0.0.0.4

Lunalicious:
- Better Localisation System to not have all the data in Memory all the time but only the relevant language
- Small changes to json config file

I'm still not 100% happy with the localisation solution but I guess it's fine overall


Lunalicious:
I'll tinker a bit more with Localisation so.. closing this for now c:

PR #3533, opened 2024-05-23
Echoglossian v3.2.x

lokinmodar:
New functions added!

- Deepl Translation Engine support thanks to @samulopez

To enable it check the plugin configuration window using /eglo command and go to the Translation Engines tab!


philpax (approved):
Code looks fine, but I hope there aren't any adverse consequences from masquerading as the DeepL app. Is there no real API?

lokinmodar:
> Code looks fine, but I hope there aren't any adverse consequences from masquerading as the DeepL app. Is there no real API?

Yes there is. There is an option that enables the use of your own API key (either free tier or paid ones) :)

PR #3534, opened 2024-05-23
Infinite Roleplay

serifas:


serifas:
please pause review of the plugin. I am going to push a new commit as I just fixed a major issue

serifas:
okay I updated the source. It should be thread safe now. Sorry about that.

PR #3542, opened 2024-05-25
ChatTwo 1.25.4

Infiziert90:
[nofranz]


reiichi001 (approved):

sersorrel:
fyi: I believe `nofranz` does not work unless it is the very first thing in the PR text, i.e. here it did not work because there's a `[` before it

see https://github.com/goatcorp/XLWebServices/blob/b40c5aac30b21e1aa16bc66cbae85ec1df24ecd2/XLWebServices/Controllers/PlogonController.cs#L242

PR #3550, opened 2024-05-27
[Testing] Eventy 0.0.0.1

Infiziert90:
**Never forget a seasonal event ever again**

Simple but useful calender
- Current events displayed within a server bar entry for quick viewing
- Browse through current, upcoming and previously held events
- Special dates for Liveletters and other important occasions are highlighted


Infiziert90:
Doesn't require any ingame functions, so could safely update to apiX already

goaaats:
bleatbot, rebuild

lmcintyre (approved):

PR #3582, opened 2024-06-03
SubmarineTracker 1.8.5.0

Infiziert90:
- This update brings with it a migration from .json files to SQLite
  - Each FC will now be identified by it's unique ID and is no longer bound to a single character. 
  - All data migration will occur in the background after logging in.
  - A notification from the plugin will provide more detailed information regarding aforementioned aspects.
  - Additionally a variety of fixes have been made as part of the migration process.

**Important Note**
While the migration is taking place, information regarding your submarines will be unavailable and won't display until the process has finished.

### DIFF
--- a/testing/live/SubmarineTracker/manifest.toml
+++ b/testing/live/SubmarineTracker/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Infiziert90/SubmarineTracker.git"
-commit = "87021188526b712c175f9db8e8bdace616777e9c"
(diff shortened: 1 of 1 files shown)


Infiziert90:
Real diff from stable to testing: (1882 lines)
https://github.com/Infiziert90/SubmarineTracker/compare/1e91d880929656c485da8447fc66e61309e42c06..87021188526b712c175f9db8e8bdace616777e9c

philpax (approved):
good luck with the migration

PR #3592, opened 2024-06-07
[testing] Dynamic UI Scaling 0.0.0.1

HybridEidolon:
Initial testing version of this plugin.

Dynamic UI Scaling is a minimalist modification of the built-in high res scaling behavior that makes the UI scale automatically relative to the base resolution. So long as you use the same aspect ratio, the scale and placement of your addon windows will be proportional regardless of resolution.


goaaats:
bleatbot, approve

reiichi001:

reiichi001:
bleatbot, rebuild

HybridEidolon:
I'll update this PR to use the revision compatible with 7.0 tomorrow

HybridEidolon:
> I'm on 3440x1440 and the minimum scale the plugin allows (0.5) is still 150-200% of the original scale. It would be nice to at least be able to scale down until you're back at the original scale. From a user perspective, I would've assumed that a value of 1 would mean "100%" (original), but that results in 400% scale or even higher instead.

This is intended - "100%" in this plugin means "scale as if the screen is 1280x720", like how WoW handles the UI scale factor. The idea is that all elements on the screen retain their relative size regardless of resolution (assuming 16:9).

> Disabling the plugin does not restore the original scale. I'll have to restart my game if I want to disable or uninstall the plugin to make everything look "normal" again.

Right, it probably needs to invoke the un-hooked scale set function. I can restore it to 100%, but with the plugin disabled setting the HD scale factor in the system preferences should also fix it.

philpax (approved):

philpax:
Thanks for testing, @haroldiedema! Those aren't blocking issues, so I'm happy to proceed; clarifying the first issue in the UI and fixing the second issue would be welcome, but those can be done in a separate issue.

---

PR #3595, opened 2024-06-09
[Testing] Who's Talking 0.6.10.0

sersorrel:
Fixes logspam in DRS, hopefully.

I've not yet checked that this doesn't break voice indicators in regular alliance raids, hence the push to testing rather than stable.


goaaats:
bleatbot, rebuild

reiichi001 (approved):

PR #3599, opened 2024-06-10
ChatTwo 1.26.0

Infiziert90:
- Replace all sigs with CS version
- Fix flags not working in reply mode
- Force the main chat log to be always open
- Prevent emotes inside auto translation


goaaats:
bleatbot, rebuild

reiichi001 (approved):

PR #3602, opened 2024-06-10
ChatTwo 1.26.2

Infiziert90:
- Fix tell reply hotkey not working
- Save the FFXIV font to disk instead of redownloading it on every startup
  - This should prevent errors if your DNS is giving up
- Make Linkshells appear in selection again


reiichi001 (approved):

reiichi001:
bleatbot, rebuild

PR #3606, opened 2024-06-11
SubmarineTracker 1.9.0.3

Infiziert90:
- Prevent migration from failing if very old data is around


reiichi001:
bleatbot, rebuild

reiichi001 (approved):

PR #3620, opened 2024-06-16
[Testing] BeegWindow

ArcaneDisgea:
Adding the new plugin BeegWindow to testing. This plugin will automatically maximize the window on load. I have tested with it enabled in full screen and borderless windowed and it does nothing when those options are selected as far as I can tell.

This plugin exists because of the few users like myself who play in windowed rather than borderless windowed.


reiichi001:
bleatbot, approve

goaaats:
bleatbot, rebuild

philpax (approved):

PR #3627, opened 2024-06-17
SubmarineTracker 1.9.1.0

Infiziert90:
- This update brings with it a migration from .json files to SQLite
  - Each FC will now be identified by it's unique ID and is no longer bound to a single character.
  - All data migration will occur in the background after logging in.
  - A notification from the plugin will provide more detailed information regarding aforementioned aspects.
  - Additionally a variety of fixes have been made as part of the migration process.

**Important Note**
While the migration is taking place, information regarding your submarines will be unavailable and won't display until the process has finished.

### DIFF
--- a/stable/SubmarineTracker/manifest.toml
+++ b/stable/SubmarineTracker/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Infiziert90/SubmarineTracker.git"
-commit = "1e91d880929656c485da8447fc66e61309e42c06"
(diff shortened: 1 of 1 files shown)


Infiziert90:
Just a version bump from testing -> stable
Diff: https://github.com/Infiziert90/SubmarineTracker/compare/44912add32c100b2908d3560d7b76abe37ad7339..593c031474d74fefce8760c083034f2c126e34c3

reiichi001 (approved):

PR #3638, opened 2024-06-23
[stable] WorkshopOptimizerPlugin - Update images to match max resolution.

belzaru17:


philpax:
@goaaats this PR just updates images in this repository and nothing else, what is the expected behaviour?

belzaru17:
The current images are too big and give an error on loading (in the log). This resizes the images so they don't exceed the max allowed resoluition,  Example error:
17:31:29.619 ERR | Plugin image3 for WorkshopOptimizerPlugin at queue was larger than the maximum allowed resolution (1230x699 > 730x380).

PR #3643, opened 2024-06-28
Mini 2.1.0.1 and NotificationMaster 2.1.1.3

Limiana:
- Api 10 update


Limiana:
will need versions of WindowsDesktop updated in plogon for this and NotificationMaster please

goaaats:
Does that work with the built-in Dalamud runtime? We are still on 8.0.0, not 8.0.6, though it might be a good idea to upgrade anyway.

Limiana:
Yeah, it works just fine
When you pull .NET sdk image with `8.0` tag, Microsoft gives you latest minor version (which is currently 8.0.302) that contains .NET 8.0.6 and it also expects 8.0.6 WindowsDesktop, but it references only 8.0.0 as specified in configuration

PR #3664, opened 2024-07-01
Hybrid Camera 1.3.0.0

Drahsid:
API 10 update. Additionally, this adjusts the plugin so that it passes all validation issues which it previously failed.
On top of this, DrahsidLib was updated to API 10.

There is some extra code here relating to controller mode and tests/reverse engineering on movement from 6.x, but none of it is enabled.


goaaats:
bleatbot, rebuild

reiichi001:
bleatbot, rebuild

reiichi001 (approved):

PR #3709, opened 2024-07-02
Update Orange Guidance Tomestone to 1.8.1

anna-is-cute:
nofranz

There will be a larger PR later with most of my plugins batched, but I wanted to get this one out first.


anna-is-cute:
Fixed a URL being wrong from testing. By the way, the C# changed is small or maybe medium. Most of those lines come from server YAML files.

reiichi001:
> There will be a larger PR later with most of my plugins batched
It would be preferred if you could submit them individually. I know you have a lot of plugins, but this greatly simplifies review complexity.

reiichi001 (approved):

PR #3757, opened 2024-07-03
[Testing] Absolute Roleplay

serifas:
I am changing the previous name from Infinite Roleplay to Absolute Roleplay.
Everything is updated to ApiX and works great from my testing sofar


reiichi001:
I've updated the PR title to reflect the plugin name. (We prefer this because plugin merges will take the PR title for discord webhooks. It also makes it easier to tell what the PR is for)

I'll update our internal review spreadsheet with the new name.

PR #3777, opened 2024-07-03
Update NotificationMaster and Mini

Limiana:
Previous PR got removed because I renamed branch


Aireil:
bleatbot, rebuild

Limiana:
Needs either windows desktop packages to be updated in plogon or primary .net package to be pinned to the version Dalamud ships with it (later is better approach I think since you'd be building plugins against version that is actually in use), right now bot uses latest minor version of .Net
https://github.com/goatcorp/DalamudPluginsD17/pull/3643#issuecomment-2197567295

reiichi001:
bleatbot, rebuild

Limiana:
bleatbot, rebuild

Limiana:
I'll come back to this later, I have some methods to bypass winforms dependency

PR #3796, opened 2024-07-04
NoTankYou v7.0.0.0

MidoriKami:
Updated for Dawntrail

New Party List Overlay, now uses completely native ui nodes, instead of ImGui Overlay
New Banner List Overlay, now uses completely native ui nodes, instead of ImGui Overlay
New Blacklist System, and new Zone Selector

Some minor upgrades to core systems

New Modules will be coming in the near future, look forwards to them


MidoriKami:
It's not ready yet.

PR #3805, opened 2024-07-04
[Testing] RandomClassSelector

Aida-Enna:
Initial release - You type /rndc and it selects a class at random based on your specified options, and optionally switches to it's matching gearset (currently matches by either short name or long name of class).

https://github.com/goatcorp/DalamudPluginsD17/assets/5483362/e6a86d5a-87e1-4a73-9dae-730b44ea8c3a

Configurable options are as follows:
image

I envision people using it to decide what class to do roulettes as, level up for the day, do quests on, etc.


philpax:
Code looks good, but it doesn't seem like Veda's used for anything. I'd suggest removing that as it has some potentially unsafe functionality.

PR #3840, opened 2024-07-06
TargetLines 1.6.1.0

Drahsid:
Fixed a null reference exception in certain unnatural circumstances, minor improvements to the performance of fancy lines when there are a very high number of lines, and added option to filter party/alliance.


reiichi001:
bleatbot, rebuild

reiichi001 (approved):

PR #3846, opened 2024-07-06
Tracky 1.5.2.1

Infiziert90:
- Improve the tracked data for revisits [upload only]
  - The nature of this data makes it only useful in a bigger set, that's why it won't be saved locally


KazWolfe:
bleatbot, rebuild

reiichi001 (approved):

PR #3911, opened 2024-07-10
TitleEdit 2.3.0.0

RokasKil:
- Updated for Dawntrail
 - Moved some old stuff onto CS 
 - Changed the logic a bit since Dawntrail title menu operates differently and was messing stuff up


lmcintyre:
Note that this is Title Edit's adoption, which has been discussed and is good with me.

reiichi001 (approved):

PR #3938, opened 2024-07-12
Update Orange Guidance Tomestone to 1.8.4

anna-is-cute:
- Use a new method of managing VFX which should fix some bugs
- Fix invisible messages


goaaats:
bleatbot, rebuild

reiichi001 (approved):

PR #3988, opened 2024-07-14
Update TidyChat v.1.3.1

NadyaNayme:
Fix improved instance messaging for instances 4-6. Maybe fix materia attachment filter for certain name settings.


philpax (approved):
`// Oops`

PR #4018, opened 2024-07-16
BurntToast1.3.3

Xpahtalo:
Fix for the bug where all toasts would get blocked when a regex got save wrong. Also improved saving again.


reiichi001:
See if adding `project_path = "BurntToast"` to manifest.toml helps.

Xpahtalo:
thank you. i knew there was something like that, but what was too exhausted to figure it out at the time.

reiichi001 (approved):

PR #4044, opened 2024-07-17
[Testing] WTSync 0.8.0.0

KhloeLeclair:
This is the initial testing release for a new plugin intended to make it easier to do Wondrous Tails with your friends. This adds an additional window that displays the Wondrous Tails progress of everyone in your party, so long as they are also using this plugin.

Filter content by type and level. Content is sorted by the number of players that share it, and then by level. Custom sorting to come in a future update.

The server component is very simple and implemented using Cloudflare Workers. I might need to change that in the future, but I doubt this little plugin will get _that_ popular. Source code for that is here: https://github.com/KhloeLeclair/WTSync-Server

FINAL_FANTASY_XIV_2024-07-17_16-33-52


karashiiro:
Sorry for the delay, doing code review today

karashiiro:
Is there an alternative to sending the player's raw content ID [here](https://github.com/KhloeLeclair/WTSync/blob/620c1d0ec1832a779856160672e7cd3a53934f1c/Plugin.cs#L166)? We're trying to phase out using player content IDs in network requests, since they're immutable and can potentially be used for tracking people persistently.

KhloeLeclair:
I'm not necessarily opposed to sending something other than the content Id, but I'm not sure what would be a better alternative. I guess "Name@HomeWorld" could potentially work as long as I have a reliable way of reading the home world of all party members. The `IPartyList` service's `IPartyMember` instances have a `World` value but I assumed that was set to the current world rather than the person's home world, based on the docstring?

All that said, I'm probably over-complicating the code that actually reads the current party so any advice there would be appreciated.

---

Regarding tracking people, I'm hoping that the open nature of the server's source code relieves any questions there about tracking taking place. That, and the server **only** receiving the state of your Wondrous Tails makes it rather limited as far as being able to actually track someone with it. If there are any changes I could make to the server to make it more open, I'd be happy to do that as well. Maybe I should see about using GitHub Actions to deploy updates so there are publicly available logs.

karashiiro:
Sorry, I wasn't clear - I meant that players can use it for tracking each other, and the fact that party membership can't be enforced at the server level also means that it's possible for anyone to plug in an arbitrary CID to determine if someone is running the plugin or not.

I brought this to the rest of the approvers for discussions and one possible alternative we came up with is to use a hash as the database ID - each party has a unique ID associated with it that gets sent to all party members (exposed [here](https://github.com/goatcorp/Dalamud/blob/fc7b1f222d82f8a96f1241c0b295ea537870d762/Dalamud/Game/ClientState/Party/PartyList.cs#L60) in Dalamud), so you could create a hash of `"PlayerId.PartyId"` and use that as the database key instead. In addition, it might be nice to purge data after a certain amount of time, but that's a good idea for cost reasons regardless of this.

KhloeLeclair:
I'd rather not use a party-specific identifier if that can be avoided. The server not being aware of party membership was a deliberate decision related to costs.

Specifically, the server is implemented using Cloudflare Workers, which bill by number of requests. Currently, WTSync is designed so it only sends updates to the server at login/plugin-load, and when your Wondrous Tails state changes. It's also debounced so if the state changes multiple times within a few seconds, it only sends one update.

If I used a party Id as part of an identifier, it would represent a potentially substantial increase in the volume of requests since I'd need to send a new update every time the party Id changes. Players join and leave parties with much greater frequency than they login, complete WT duties, or interact with their WT books so I'd expect it to be a fairly significant request volume.

Without having the plugin live, I can't predict the volume of traffic I would be getting, so I prefer to err on the side of caution as much as possible.

I'm probably thinking about it too hard, but my day job involves very large traffic volumes so it's hard not to think about things in those terms.

---

If this is a blocker I can look into hosting the server component with some other provider where request volume is not billed directly, but getting set up with a cheap VPS will probably take me several days at least. I am not good at ops, and also don't have a ton of spare cash to throw at a hobby project like this plugin, which is what made me want to try something like the Workers platform so I didn't have to worry about that sort of thing.

karashiiro:
Workers offers 100k requests per day on the free tier, which is a very low limit, but it's not too low for this, I think - party IDs should remain the same for rejoins, so if I understand correctly:

If we say 8k users log in with this, that's already 8k updates, and if they all join full parties with other plugin users that's another 28k requests (first player makes 7 requests as others join; second player makes 6, etc... makes 28, times 1k parties).

If a player leaves and rejoins, there's still another 8 requests, as we go through a couple more state transitions:
1. 8 players -> 7 players (no new players, no new requests)
2. 7 players -> 8 players (one new player, 8 new requests)

Following the same logic, 16 requests happen if players swap between two parties.

Compare that to using party IDs as salts: the setup is all the same, but there's a 1 request difference on rejoins:
1. 8 players -> 7 players (no new players, no new requests)
2. 7 players -> 8 players (one new player, 9 new requests as the joiner does another update first)

18 requests now happen if players swap between two parties, which isn't much worse.

The bigger issue is that this introduces a race condition where party members can request states before a player has sent their current state, but you can mitigate that with either WebSocket to control sequencing (which you already seem to have) or by optimistically delaying the fetch request until the new party member is most likely to have already sent their update if they have the plugin installed. In any case, the request count shouldn't be substantially different between the two approaches.

---

To make things safer, a couple potential optimizations here regardless of the approach taken are caching by player CID (no party ID) locally to handle rejoins and moving between parties with the same players, and to defer sending requests until the first time a party is joined on a particular character (also need to handle character changes) - it's safe to assume that not every login is going to involve joining a party.

For the party ID approach specifically, you can also improve that by not resending WT state if it has already been sent for that party and hasn't changed.

KhloeLeclair:
Alright then. Well, I'll implement a contentId + partyId hash for player identification for now and worry about server capacity once the plugin is approved.

This shouldn't require many changes to the client code, outside of how I'm determining party member IDs and some modified logic for when it sends updates.

Were there any other things of concern in the code I should address while I'm at it?

karashiiro:
Nothing else that I noticed, everything else looked good to me 👍

KhloeLeclair:
So, after a bit of looking into this, I've run into an issue with PartyId. Specifically: cross-world parties. Is there a way to get a stable shared Id for a cross-world party? If so, I haven't seen one.

I've been using `FFXIVClientStructs.FFXIV.Client.UI.Info.InfoProxyCrossRealm` to read the state of a cross-world party, but unless I've missed it, that doesn't contain any party Id. Given how half-baked cross-world parties are, I would not be surprised if they don't have an Id like a standard party would.

karashiiro:
Looking at ClientStructs, there's a party struct called [Group](https://github.com/aers/FFXIVClientStructs/blob/a0bc6edcd7535330d52fe33dcf453bc09fcc51fd/FFXIVClientStructs/FFXIV/Client/Game/Group/GroupManager.cs#L25-L26) which appears to have two party IDs, it might be that one of them is same-world and one is cross-world?

karashiiro:
IIRC InfoProxy objects contain data that the UI reads from (it's like the Model or ViewModel in an MVC architecture), so it's likely that doesn't contain much data that people aren't going to see.

KhloeLeclair:
Neither ID gets set in a cross-world party, I'm afraid. Both `PartyId` and `PartyId_2` are just `0`. I've tested both with and without using party finder to set up the party, in case that would matter but it doesn't.

karashiiro:
Just had a chance to try it myself - I think they're `0` until you actually have party members - I set up a command to print them both:

```csharp
public unsafe void GetPartyIds(string command = "", string args = "")
{
    var group = GroupManager.Instance()->GetGroup();
    this.chat.Print(new StringBuilder()
        // this was just to see if this unmapped field had something useful
        .AppendLine($"Group.0x6580: {Marshal.ReadInt64((nint)group, 0x6580)}")
        .AppendLine($"Group.PartyId: {group->PartyId}")
        .AppendLine($"Group.PartyId_2: {group->PartyId_2}")
        .ToString());
}
```

Running this in an empty PF (both with and without crossworld enabled) resulted in both being `0`, but during a crossworld Duty Finder raid they both had values.

KhloeLeclair:
Indeed, once you're actually in instanced content you have a real party so there's a real party ID to be found. But when you aren't in an instance, you only have the cross-world party and not a real party, and there is no way to have a real party. If I was to rely on this as a source of a synchronized ID, at best this means I could only start syncing data once everyone in a cross-world party had done a duty together and that just doesn't seem at all reasonable to me.

As we have yet to discover any sort of ID for a cross-world party, let alone a stable ID, do you or the other members of the committee have any further guidance / ideas on how to handle this? I know you have concerns about using content ids, but I absolutely don't want to exclude people in cross-world parties from benefiting. Some of the friends I've been having help me test this use cross-world exclusively pretty much and I can't imagine it'd be that rare in the community.

That said my only other immediate thought is "Name@HomeWorld" and that's not exactly better from a standpoint of not letting users be identified.

I suppose it could use the party Id for people in a normal party, and some alternative fake party Id for a cross-world party? But that'd likely be leaving the content Ids of people in cross-world parties exposed unless you've got a better idea than 'use a constant'.

(Honestly Square could you just please make cross-world parties stop sucking already, it's been years...)

---

As an aside, I just want to say thank you for your time going back and forth on this.

karashiiro:
> Indeed, once you're actually in instanced content you have a real party so there's a real party ID to be found. But when you aren't in an instance, you only have the cross-world party and not a real party, and there is no way to have a real party. If I was to rely on this as a source of a synchronized ID, at best this means I could only start syncing data once everyone in a cross-world party had done a duty together and that just doesn't seem at all reasonable to me.

I was seeing the empty IDs with an empty same-world PF too, do you mean same-world and crossworld PFs behave differently after players have joined the party, but before joining an instance? I figured that as soon as someone else joined either party type those would be populated, and it just appeared like they were different since I was mostly testing with empty parties.

> That said my only other immediate thought is "Name@HomeWorld" and that's not exactly better from a standpoint of not letting users be identified.

This might actually be fine, even if it's not ideal - player names can be changed, though it isn't trivial to do so. The concern with CIDs specifically is that they can never be changed, so they can be used to circumvent anything someone could try to do to break links to themselves if they needed to.

Just using name/world would still mean having a way to see if non-party members are using the plugin, which isn't great, but it might be possible to at least minimize the risk from that by having the same server response for people who aren't using the plugin (missing DB entry which I believe currently returns an error) as for people without a WT book. That way, it's at least ambiguous when looking up plugin-users who aren't actively doing WT. I'll have to check with the other approvers again, but I think that might be enough.

KhloeLeclair:
Alright, that sounds positive. Switching to using the user's name and homeworld would require pretty minimal work. The server already uses a string for the Id, so I'd just need to change how it validates them effectively.

I can also change the HTTP responses so all requested users get a response, even if the response is just that they don't have an active book.

On the top of users who aren't active, I could try also adding a logout event that fires off a quick HTTP request to delete the user's state in the server, so it'll only have records for users while they have the game open. I'm not sure how effective that'd be though, since I don't know how quickly Dalamud needs to wrap things up when users exit the game.

karashiiro:
Adding a logout handler won't be completely reliable, but it should be useful most of the time - unless someone uses the "kill game" option, there should be time for a proper logout event.

KhloeLeclair:
Alright, I've changed to using a sha256 hash of "Name@HomeWorld" to identify players. This is a fairly simple change, though it necessitated the switch to GroupManager from IPartyList to ensure I was getting HomeWorld and not just the player's current world. (This may have been a docstring error but I wasn't able to test for sure and I wanted to make certain.)

Also added a logout event handler that tries to fire off a request to clear existing data. Hopefully this helps address the committee's privacy concerns.

karashiiro (approved):

PR #4102, opened 2024-07-21
[Testing] Tidy Chat 2.0.0

NadyaNayme:
Rewritten internals to make adding new filters faster and easier. Very few user facing changes - the largest one being the removal of 'Hide' settings. All settings are now 'Show' - please adjust your settings as necessary.

I expect some filters may not work as expected - however troubleshooting is now significantly easier as `/xllog` will report what filters a message did or did not pass and when Debug Mode is enabled in the settings you can test using /echo as well to confirm if a rule is being matched or not.


philpax:
This is going to take a while

philpax (approved):

PR #4130, opened 2024-07-22
Update Orange Guidance Tomestone to 1.9.0

anna-is-cute:
nofranz

This update adds support for the "phantom" players performing an emote you can see in some Souls games when viewing a message.

To facilitate this, the plugin sends the equipment model information, weapon model information, and full customise array of the player to the server, where it is stored with the message data. Deleting the message (or the account associated with it) deletes the data. This is explained fully in a new data/privacy section in the settings.

As before, accounts have no PII anyway, so this cannot be used to tie any identity to any message.

I have tested spawning these phantom characters with random data for the customise array and all model information, and due to the normalise function being called, I cannot cause the game to crash. The characters are spawned and explicitly marked as BattleNpcs and set untargetable.

As far as I understand it, this should all be above board and safe.

### DIFF
--- a/stable/OrangeGuidanceTomestone/manifest.toml
+++ b/stable/OrangeGuidanceTomestone/manifest.toml
@@ -1,9 +1,10 @@
 [plugin]
 repository = 'https://git.anna.lgbt/anna/OrangeGuidanceTomestone.git'
-commit = 'c65ff3837854c2494344d6e6130f3111dca0402a'
+commit = 'e789a2e7666dcab9ae098018bc86e1e615dcb13d'
 owners = [ 'anna-is-cute' ]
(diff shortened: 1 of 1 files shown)


philpax (approved):
Looks good! There was some concern raised amongst the review team that spawning a battle character might pose some risk, but it looks like you've tested it sufficiently. I trust you'll keep an eye on this in future :slightly_smiling_face: 

Cool feature, though. Looking forward to seeing what people think!

anna-is-cute:
> spawning a battle character might pose some risk

Not as far as I can tell, but if there's any feedback in the Discord about it being a problem, let me know and I'll try to fix it.

PR #4149, opened 2024-07-24
Update Orange Guidance Tomestone to 1.10.0

anna-is-cute:
Fix face emotes causing phantoms to A-pose, and add support for changing vfx opacity and intensity.


philpax (approved):
Figured you'd want to know that people are really enjoying the ghosts, btw. Nice work!

PR #4158, opened 2024-07-24
DailyDuty v5.0.1.0

MidoriKami:
Normal Raids now work


philpax (approved):
burying the actual change in the dependency 😆

MidoriKami:
Yeah, it was a mistake when updating DailyDuty to the new version of KamiLib, there were no more weekly limited normal raid drops at the time, so had no way to verify the module.

PR #4163, opened 2024-07-24
[Testing] Glamaholic 1.10.1

caitlyn-gg:
- Added configuration migration to keep previously saved dyes
- Added more verbose logging to EC imports


caitlyn-gg:
PR updated, I had wrongly assumed that Dalamud automatically created a config file for plugins if one did not already exist. The update checks that a config file exists before attempting to read it in manually.

reiichi001 (approved):

PR #4195, opened 2024-07-28
MacroMate 1.0.14.0

grittyfrog:
Changes:

- Newly created macros will now be 'Always Linked' by default (previously all conditions were added)
- Added a new button to add all conditions
- Improved Help window with importable examples and formatting fixes

Bugfixes:

- Fix initial size of macro window being too small
- Fix macro name width with high global font scale
- Fix padding with high global font scale
- Fix condition editor button alignment


philpax (approved):
thought that tutorial was a GPT prompt at first

PR #4280, opened 2024-08-03
Chatter Updated to v1.2.0--events

jlkeesey:
Added support for timed events.

* Timed events are groups that can be started from the main panel and will automatically stop or deactivate when the time runs out. This can be used for short events like Eternal Bonding ceremonies.

* The /chatter command now opens the main window which has controls for running Chatter. The configuration window command is now /chattercfg. You can also open the config from the new main window.


reiichi001:
Please make sure to also update your csproj
https://github.com/jlkeesey/Chatter/blob/main/Chatter/Chatter.csproj#L7

reiichi001 (approved):

PR #4375, opened 2024-08-13
[Testing] Item Use 0.9.0.0

PunishedPineapple:
This is a simple new plugin that shows icons on the top right of your item information to indicate when an item is used for grand company deliveries and levequests, as well as to make it faster to see when they are crafting materials or aquarium fish than looking through the item description.  You can set which categories of items you want indicated, and you can have them be either separate icons for each category, or one single icon just to show that the item is "useful".  The primary use case is quickly seeing which gear/food/fish/mats have value, and which are likely to just be desynth fodder or vendor trash without having to look anything up outside of the game.

If you run into any general issues or find that an item is not being flagged correctly, please create an issue on this plugin's Github repo.

Current Limitations:
- You have to manually set your GC in the configuration window if you want your GC icon to be the Adders or Immortal Flames.
- The icon for levequests is a die instead of the leve plate icon (dice seemed the closest thing available to cards).

Planned and Potential Features:
- Showing on which jobs using an equipment coffer will produce an item used for levequests or GC turn-ins.  This requires a lot of manual data entry, so when it is added, it will not be available for all coffers at once.
- Highlighting the "Crafting Material" and "[Suitable for display in aquariums...]" text in different colors as an alternative to icons.


philpax:
I notice that `URLLink` will call `Process.Start` on whatever "URL" is passed to it, which is quite scary. I can't find any use of the function within the repository, but I'd suggest validating the URL being passed into the function to make sure it's a valid URL and can't be used maliciously.

philpax (approved):

philpax:
@sersorrel tested this and had this to say:
> ItemUse loads, it puts the icons in the tooltip like it says it will, the settings work fine
> it produces a couple of **exceptionally long** xllog messages on load ("Loaded GC item IDs:" etc.)

Consider cleaning that up at your convenience, but that's good enough to merge now!

PunishedPineapple:
> I notice that `URLLink` will call `Process.Start` on whatever "URL" is passed to it, which is quite scary. I can't find any use of the function within the repository, but I'd suggest validating the URL being passed into the function to make sure it's a valid URL and can't be used maliciously.

@philpax To be honest, the file that this is in was just copy-pasted from a previous plugin as-is and did not remove any unused functions.  I will take it out for the next release, since it has no use here really, although I should go back and fix it for the other plugins that *do* use it at some point.

As for the long log messages, they should only be present on verbose and debug log levels, but I can replace some of them with other ways of seeing what got loaded.

Anyway, thank you (and anyone else that looked over it) for your time reviewing this plugin.

PunishedPineapple:
Actually @philpax, sorry to ping you again, but looking at the URL issue, I see that Dalamud provides a function to open links (`Dalamud.Utility.Util.OpenLink`) and I was just going to switch to calling that.  However, it seems it also naively just executes the provided "URL", so idk if anyone wants to add validation to that function in Dalamud or not.

philpax:
Thanks, I've mentioned it in Discord. That should probably be fixed!

KazWolfe:
> However, it seems it also naively just executes the provided "URL", so idk if anyone wants to add validation to that function in Dalamud or not.

This has been known and will be addressed in a future release of Dalamud. For now, it's up to the plugin developer to handle any link resolution and cleanup accordingly. Unfortunately there's no simple way to tell Windows "open this in the default browser specifically", so we're left dealing with this.

PR #4379, opened 2024-08-13
Umbra 2.2.19

haroldiedema:
# Umbra 2.2.19

## New Additions

- Added conditional visibility options to the Auxiliary Bar.
- Added a "Duty Recorder" indicator widget that shows a simple icon whenever the duty recorder is active. This widget is not interactable, and only exists to show the indication of the recorder being active, similar to how the mail indicator widget works.
- Added a "Text Size" options to most widgets. These allow you to configure the font sizes on individual widgets using a single label, or the double label displays respectively.
- Added a "Maximum text width" option to most widgets which allows text - and thus the widget - to grow up until a certain size, after which the text will get truncated and appended with an ellipsis symbol.

## Fixes & Improvements

- Fixed the World Name widget not honoring the "Hide on home world" option.
- Fixed the tooltip not working on the "Item Button" widget.
- Replace "Aufgaben" with "Inhalten" in German translations (By [Bloodsoul](https://github.com/Bloodsoul)).

### DIFF
--- a/stable/Umbra/manifest.toml
+++ b/stable/Umbra/manifest.toml
@@ -1,41 +1,23 @@
 [plugin]
 repository = "https://github.com/una-xiv/umbra.git"
-commit = "7b141da8108a21375e19a3701e4025039a3141b1"
+commit = "09ce2230bc754f57dbe8439bcfa3fe1fd04f4b10"
 owners = ["haroldiedema"]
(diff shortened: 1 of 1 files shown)


philpax (approved):
I get the feeling we should be merging these faster so that the diffs don't grow out of control

PR #4380, opened 2024-08-14
DailyDuty v5.0.3.5

MidoriKami:
nofranz

Correctly set thing to null, so the null checks actually work.


MidoriKami:
bleatbot, rebuild

philpax (approved):
broken = null;

PR #4424, opened 2024-08-18
[Testing] Item Use 0.9.2.0

PunishedPineapple:
- Icons for coffer jobs have been moved into the item description.
- Added optional highlighting of the text "Crafting Material" and "[Suitable for display in aquariums tier X and higher.]" in item descriptions.
- The player's grand company is now determined automatically when displaying the GC icon.


PunishedPineapple:
Sorry for another non-trivial update to this so soon.  It should be the last time unless testing comes across any issues.

reiichi001 (approved):

PR #4440, opened 2024-08-20
DeathRoll 1.5.3.0

Infiziert90:
- Complete redo roll parsing 
- Fix "Draw Both On Start" bugging out after first card
- Fix "21 Blackjack Behaviour" not working correctly
- Potentially fix a bug around draw order

- Add a very early prototype of peggles like game


philpax (approved):
yeah okay why not

PR #4476, opened 2024-08-26
DailyDuty v5.1.0.0

MidoriKami:
New Configuration System for Todo List and Timers, you now have more style control than ever before.
**Unfortunately that means your existing timers and todo configurations are going to be nuked**

Probably fixes Treasure Map Module

Allow overlays in PvP

### DIFF
--- a/stable/DailyDuty/manifest.toml
+++ b/stable/DailyDuty/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/DailyDuty.git"
-commit = "711ee430c056b80301316ee61dd900dab783920d"
+commit = "64958c035d9d5d7b4be51a58840c7943c042295c"
 owners = ["MidoriKami"]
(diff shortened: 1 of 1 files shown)


philpax:
I'm going to hold off on merging this until you confirm that you're ready for wiping peoples' existing data and the reaction that will occur in response.

MidoriKami:
> I'm going to hold off on merging this until you confirm that you're ready for wiping peoples' existing data and the reaction that will occur in response.

It's only the layout information that's getting wiped, DailyDuty has individual config files for each sub system to make data resets not impact the rest of the plugin. Sometimes this does go wrong and there are bugs but generally this has been fairly painless.

The new options provided address sever user complaints about not being able to configure certain aspects of the overlay elements.

Additionally this isn't mentioned in the change log but I have removed the requirement for high resolution textures and have adjusted KamiToolKit to correctly render parts regardless of user setting.

PR #4518, opened 2024-08-31
[Testing] Altoholic v0.0.0.11

Sohtoren:
* Migrate database provider from litedb to sqlite
  * Data should migrate without issue however in the case it doesn't, disable the plugin and manually delete the current database file in the plugin config folder before re-enabling.


philpax:
Will users lose their data if the migration fails?

Sohtoren:
> Will users lose their data if the migration fails?

Not per se, the migration will try again the next time the plugin is loaded.

It's mostly for the unused currency history, characters data will be the same next login on each anyway.

philpax (approved):

PR #4542, opened 2024-09-05
Delete .idea directory

reiichi001:
This isn't supposed to be here.


wolfcomp:
should add `.idea/` to a `.gitignore` file to not have this get pushed up ever

PR #4573, opened 2024-09-14
[Testing] DragoonMayCry v0.12.0.2

Felscream:
- Sound files are now downloaded separately from https://github.com/Felscream/DragoonMayCry/releases/tag/v0.12.0.0 instead of being bundled directly with the plugin.
- Files are extracted into the DragoonMayCry's pluginConfigs directory.
- Each time the plugin is loaded, it will do a file integrity check on the additional assets.
- If the integrity check fails, the user is notified, existing assets are deleted  and new ones are downloaded from the link above.
- Added more sound effects and dynamic background music selectable on a job-per-job basis.
- Dynamic background music is opt-in, and only playable inside non PvP instances. 
- If the user configured a job to use dynamic background music, it will mute the game's BGM when they enter an instance with that job.
- The game BGM state is restored when they leave the instance, or choose a job inside the instance which doesn't have dynamic music enabled.
- Dynamic BGM files are cached in memory when the user enters an instance. They are cleared only when the plugin is disabled.
- Lots of duplicated code in the FSM namespace

### DIFF
--- a/testing/live/DragoonMayCry/manifest.toml
+++ b/testing/live/DragoonMayCry/manifest.toml
@@ -1,10 +1,19 @@
 [plugin]
 repository = "https://github.com/Felscream/DragoonMayCry.git"
-commit = "8f517002fe4ef5e65b6bb892192bd606d980479d"
+commit = "02b60559f5d8b1e19d4ce083175d344f8a16e18e"
 owners = ["Felscream"]
 project_path = "DragoonMayCry"
 changelog = '''
-v0.7.5
+- Added announcers and dynamic background music selectable on a job per job basis
+- Announcers :
+    - Default DmC5 (new default)
+    - DmC5 Balrog VA / Michael Schwalbe
+    - Nico
+    - Morisson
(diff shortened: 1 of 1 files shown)


gnargle:
Done some initial testing. New features look good and I'm not getting any major bugs or issues, nor any errors/warnings in xllog.
I did notice some functionality bugs with the new voices available:
* Nico nor Morrison's voice do not function for D and C ranks, instead defaulting to the DMCV voice. IDK if this is because they doesn't have lines for D and C but from hitting the sample button a few times, it sounded like they do.
* The only way I could hear the bgm without cranking the volume to max was to turn "apply game volume..." off. IMO remove the option entirely and have the volume be settable from the plugin only, else it adds too many variables.

Felscream:
For Morrison and Nico, that's intended. The original game doesn't provide lines for D and C, and defaults to the original announcer (I checked on youtube).
As for the BGM volume, maybe I'll default it to off later but I personally don't like loud BGM, that's why it's there at least to try and keep it in line with the rest of the game's volume on my personnal setup

philpax (approved):

PR #4584, opened 2024-09-17
Hybrid Camera 1.4.0 (Testing)

Drahsid:
Update to hybrid camera I was working on around the end of Endwalker. 
This adds controller support, and simplifies a lot of the implementation, allowing me to remove configuration options.
This might be illegal, since it directly hooks the movement code to adjust the behavior and switch from standard/legacy movement. I am putting this version in testing for now just in case there are some side effects I did not catch, though I did personally test to ensure various things still work (knockbacks, grabs, sliding floors, being frozen/unable to move, etc.)


Drahsid:
I ended up doing some more work on this. One more option has been added to make standard-style backpedaling move at full speed. This mechanically has no difference in comparison to legacy-style backpedaling (besides lower acceleration,) but might look strange to other players.

philpax:
Hi, apologies for the delay on this; we've been deliberating over the potentially illegal behaviour. We hope to have movement on this soon.

Drahsid:
No worries, I thought as much.
I think if anything here is specifically illegal, it is the full-speed standard style backpedaling. Even though it has the same max speed and lower acceleration than legacy, it is pretty obvious when a player is using it, even the movement latency doesn't mask it.

lmcintyre:
Hi! We've taken an extra close look at this and have deemed the "Fullspeed backpedal" too obvious to allow. To other players, it looks like a speedhack (and essentially is, considering the movement you're doing, backing up without turning, is not intended to be performed at that speed). The other smart strafe features are fine.

Drahsid:
I've gone ahead and removed the fullspeed backpedal option.

reiichi001 (approved):

PR #4594, opened 2024-09-18
[Testing] DragoonMayCry v0.12.2.1

Felscream:
- Reworked the job configuration UI
- Added the ability to disable the plugin for specific jobs
- Bugfixes


Felscream:
Fixed a behaviour where changing the default audio output device while the plugin was running would not redirect sound to the new device.

Felscream:
- Added a muffled effect on death
- Sometimes the dynamic BGM wouldn't transition to the "in combat" state due to an error

reiichi001 (approved):

PR #4615, opened 2024-09-25
Allagan Market 1.0.0.6

Critical-Impact:
**Fixes**
- Fixed an issue where old prices would still persist even if they were no longer applicable.
- You should no longer be notified of a previous valid undercut that is now no longer valid(notifications are queued so they are re-evalulated at the time of sending)


philpax:
bleatbot, rebuild

reiichi001 (approved):

PR #4653, opened 2024-10-05
Update EngageTimer v2.4.1.0

xorus:
- Option to hide the stopwatch/floating window in cutscenes
- Enabling light effect animation would show an error texture and required a plugin restart to be loaded properly


goaaats:
bleatbot, rebuild

goaaats (approved):

PR #4663, opened 2024-10-10
Discord.RichPresence 2.0.5.0

reiichi001:
Now includes a setting to automatically run a Wine IPC Bridge for Linux users (if you need one).

XIV On Mac users should continue to use the Wine IPC Bridge available in the XOM program settings.

This change has no effect for Windows users of the plugin.

### DIFF
--- a/stable/Dalamud.RichPresence/manifest.toml
+++ b/stable/Dalamud.RichPresence/manifest.toml
@@ -1,11 +1,15 @@
 [plugin]
 repository = "https://github.com/reiichi001/Dalamud.RichPresence.git"
-commit = "07ba4060253c96725e4005844c777d6e60f73957"
+commit = "2c0e8a68b2117eec29daa422c81db4b00ffc62d4"
 owners = [
     "goaaats",
     "reiichi001",
 ]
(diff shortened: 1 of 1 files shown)


reiichi001:
If our build system doesn't like pulling the Wine IPC Bridge exe from Blooym's PR, then I can ship it directly or it can probably be reworked to be downloaded on demand.

Blooym:
I thought the build system could pull in files at build time but if not directly shipping would work fine too

reiichi001:
bleatbot, rebuild

reiichi001:
I'll need a Linux user to test this out. I *think* I adjusted paths correctly, but I'd like some eyes on that before its merged because I don't currently have a Linux machine ready for FFXIV.

@Blooym if you've got a moment, can you pull the artifact? (Otherwise, I can poke the linux channel for a sacrifice)

reiichi001:
I spun up a fedora machine and it worked there. This should be fine to merge now

Blooym:
Just tested and also worked fine

philpax (approved):

PR #4666, opened 2024-10-10
Allagan Tools 1.7.1.0

Critical-Impact:
This is a fairly large release, please expect some bugs
### Added
- Curated lists(build your own lists with whatever you want in them)
- Added a new menu bar to the list/craft windows
- Added more options for copying/pasting
- Craft zone, dye count, materia count, are recipes completed and remove columns added
- Dye count, materia count, are recipes completed filters added
- The dye column can be configured to show stain 1/2/both
- Added a new layout(single)
### Fixed
- The second dye will now track (thanks emyxiv)
- Craft columns could not be edited
- Fixed market ordering
- The way inventory sorting is tracked should be faster

### DIFF
--- a/stable/InventoryTools/manifest.toml
+++ b/stable/InventoryTools/manifest.toml
@@ -1,15 +1,25 @@
 [plugin]
 repository = "https://github.com/Critical-Impact/InventoryTools.git"
-commit = "8b52fd6eec34bfae78388caea84b2170da838bee"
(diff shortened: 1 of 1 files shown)


Critical-Impact:
Apologies, I generally don't like to bump commits while they are being reviewed but a friend found a bug caused by a few missing services, just adds 3 lines so those services are actually loaded

reiichi001 (approved):

PR #4711, opened 2024-10-25
Mappy v3.0.3.5

MidoriKami:
nofranz

fixes area/subarea text showing in title when it shouldn't


MidoriKami:
Oh, that's actually an older KamiLib, I guess I can update it to the version that's been reviewed already.

philpax (approved):

PR #4714, opened 2024-10-27
Multi-plugin drifting: Updating 4 plugins' URLs and nothing else (except there was other stuff, teehee)

Aida-Enna:
~~Can we update 3 plugins at once? Time to find out!~~

Scratch that, we're pushing the boundaries with FOUR plugins updating at once - 3 of them just have their URLs fixed (and a version bump), but graphics config also has a small bugfix.


NotNite:
Looks like a no :p

philpax (approved):
h

PR #4725, opened 2024-10-30
Absolute Roleplay [Testing]

serifas:
New fork to clean things up


reiichi001 (approved):

MidoriKami:
Just to clarify terms, I think you mean new branch, as a new fork would be a different repository.

Forks are typically used by people other than the repository owner so they can make their own changes to the project.
Often forks are used to then contribute code back to the origin repository, but not always, sometimes someone wants to undo something that you did, or they want to take your code base in a different direction.

PR #4730, opened 2024-10-31
NoTankYou v7.1.0.2

MidoriKami:
Fix Monk Warnings appearing when too low level


MidoriKami:
Not sure why it's saying that KamiLib 5ad86f9c8a6373f6b41dea915f0adce10373d05f is new, when it is shown as reviewed already in a different PR https://github.com/goatcorp/DalamudPluginsD17/pull/4695#issuecomment-2427984512

KamiToolKit however probably does need a review, as this is the first of my plugins with KamiToolKit that's been updated. Worth noting, this specific update didn't touch KamiToolKit, so it's unchanged since last review.

philpax (approved):

PR #4797, opened 2024-11-15
CurrencyAlert v2.2.0.2

MidoriKami:
API11

I expect no issues with this one.

### DIFF
--- a/stable/CurrencyAlert/manifest.toml
+++ b/stable/CurrencyAlert/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/CurrencyAlert.git"
-commit = "3e760f6eda238b0c95099150a6d9b71f661070e8"
+commit = "b476962617cbd89d08453ecd3e4008e1e334370c"
 owners = [ "MidoriKami" ]
 project_path = "CurrencyAlert"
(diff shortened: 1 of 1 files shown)


MidoriKami:
No idea why the build is failing, log suggests some kind of internal build error with packager.

reiichi001:
bleatbot, rebuild

reiichi001:
bleatbot, rebuild

MidoriKami:
It definitely builds locally for me, I genuinely don't know what the problem could be. NoTankYou seemed to report the same issue, seems like it failed to build once after it was merged, but then seemed to build fine afterwards.

goaaats:
This must be an issue in DalamudPackager caused by the new SDK, in connection to libs. Not 100% sure how to solve it yet but I'll look at it tonight. We might have to have a permutation of Dalamud.NET.Sdk that doesn't include DalamudPackager.

MidoriKami:
I just don't know why it would work for my other plugins and not this one, I'm not aware of doing anything differently here.

MidoriKami:
bleatbot, rebuild

reiichi001:
bleatbot, rebuild

reiichi001:
bleatbot, rebuild

MidoriKami:
Removed dalamud.NET.sdk from KamiToolKit, and KamiLib.
I very much would have preferred to keep using the sdk for these projects.

reiichi001 (approved):

PR #4799, opened 2024-11-15
EzFauxHollows v1.4.0.0

MidoriKami:
API11

I expect no issues with this one.


MidoriKami:
Size is due to reformatting the plugin style.
I skipped updating this one in 7.0 when I changed the code style of all my plugins as there were no Unreal Trials available at the time.

reiichi001 (approved):

PR #4888, opened 2024-11-17
Distance 1.1.3.1

PunishedPineapple:
- Updated for Dalamud API 11 and .net 9.
- Please note that arcs (both boss aggro and custom) are not currently being drawn on-screen.  Please do not submit feedback about this known issue.  The Dalamud API function for converting world coordinates to screen coordinates is behaving differently than before patch 7.1.  Depending on whether that is intentional, a future plugin update may be required to address this once the dust has settled.  You can still enable the boss aggro distance text widget if desired.

### DIFF
--- a/stable/Distance/manifest.toml
+++ b/stable/Distance/manifest.toml
@@ -1,10 +1,11 @@
 [plugin]
 repository = "https://github.com/PunishedPineapple/Distance.git"
-commit = "8590a50e3a1bf5c3ee0afcbdd88d8aba32dce70a"
+commit = "9b73eb9d7ce41e1934085b25519dba4ce92e5ecc"
 owners = [
     "PunishedPineapple",
 ]
 project_path = "Distance"
 changelog = '''
-- Displays distance units as meters now instead of yalms in the Japanese-language client in order to match tooltips.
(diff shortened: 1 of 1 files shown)


MidoriKami:
net9 is delayed for a lil bit, API11 is net8 still

PunishedPineapple:
Understood.  I will roll back then.

Infiziert90:
The World to Screen function should work just fine, i use them in my plugins and also tested them in PixelPerfect today ... maybe because you had it set to net9 branch which was not the latest CS version? Would definitely recommend retesting with net8/stg branch

PunishedPineapple:
Dalamud isn't updating for me when set to stg, so I assumed net9 was what was getting used based on what was said in the discussion post on the Dalamud repo earlier this week.  I'm just going to draft this until Dalamud is updated for the masses.  I don't play anymore and am just too tired to try to figure out why I'm not getting stg.

reiichi001:
It's likely the beta key/kind. Both have to be set in your dalamud config. (Public testing hasn't started just yet, so the only place that's been announced is in the private dev channels. Feel free to DM me on Discord though and I can sort that out for you if you want)

reiichi001 (approved):

PR #4891, opened 2024-11-17
Allagan Tools 1.11.0.0

Critical-Impact:
### The Lumina Rewrite
- With support for API11 comes a rewrite of AT's data layer that was required due to lumina changes(how we access the game's sheet data)
- There will definitely be some broken things and changes with how the sources/uses are displayed/used in craft lists so please bear with me
- Sources and uses will generally be more detailed now and hovering over them will provide you with seal costs/rewards/etc
- Memory/load time improvements, I've seen as little as 100MB not factoring in lumina's own caches, I've seen it load in under 2-3 seconds now too
- Any bugs, please jump on the xivlauncher discord or from AT's menu hit File -> Report a issue

### DIFF
--- a/stable/InventoryTools/manifest.toml
+++ b/stable/InventoryTools/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Critical-Impact/InventoryTools.git"
-commit = "7a7e82c779959710544786d24f80bb4b75cab516"
+commit = "160d5b81de30d0fdd7ec47b191630912bd44e9fd"
 owners = [
     "Critical-Impact",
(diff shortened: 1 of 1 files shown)


Critical-Impact:
To the PAC, I do apologise in advance, obviously I'd rather not have had to rewrite the entire data layer but the way I was using lumina was wholly incompatible with how AT was built

I've tried to limit the actual changes to what was required to work with the new data layer I wrote(AllaganLib.GameData) but some changes were unavoidable

Definitely leave this for last and thank you again PAC

reiichi001 (approved):

PR #5001, opened 2024-11-21
ChatTwo 1.29.9

Infiziert90:
- Handle NBSP payloads
  - Only affects french speaking people with weird windows settings 
- Use ingame version of axis font
  - This affects the new party number symbols
  - And it won't work for the webpage version


Infiziert90:
bleatbot, rebuild

reiichi001 (approved):

PR #5005, opened 2024-11-21
[Simple Todo] - API 11

Lunalicious:
Update to API 11

This includes literally just the change from 10 to 11 in the plugin's json config and the SDK update :D

Thanks for your work <3

### DIFF
--- a/stable/SimpleTodo/manifest.toml
+++ b/stable/SimpleTodo/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://gitlab.com/moonfox-projects/simpletodo.git"
-commit = "95f85fa732f5c47206c79af96b12cf376d1e0a23"
+commit = "a75f8805e8225184e529d19b5e5958017818e5e2"
 owners = ["Lunalicious"]
 project_path = "SimpleTodo"
 changelog = "Update for API 11"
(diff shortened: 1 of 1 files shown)


philpax:
Please update the version of your plugin - happy to merge once that's done 🙂

Lunalicious:
Oooh the fourth number doesn't suffice apparently :D Well, I guess I'll just remove that one then. Sorry for the inconvenience

goaaats:
bleatbot, rebuild

reiichi001:
You're only modifying the plugin json manifest, not the csproj where it should be defined. (You don't actually need to set the AssemblyVersion, ApplicableVersion, or DalamudApiLevel in your json. DalamudPackager will take care of filling those in for you!

You'd wanna update the version number here: <https://gitlab.com/moonfox-projects/simpletodo/-/blob/master/SimpleTodo/SimpleTodo.csproj?ref_type=heads#L14>

(Looking at your csproj, you could also trim Author from your json too since you set that in the csproj)

Lunalicious:
I'm terribly sorry for the delay, I had a lot going on recently...
I removed some unneccessary stuff from the JSON but the Author was required to build it.

Thank you for your patience and your help! <3

reiichi001 (approved):

PR #5026, opened 2024-11-22
MacroMate 1.0.20.1

grittyfrog:
- Fix crash when loading Macro Mate on boot


goaaats:
bleatbot, rebuild

grittyfrog:
That diff count is definitely wrong, it's a ~45 line change!

reiichi001 (approved):

PR #5039, opened 2024-11-23
SortaKinda v2.1.0.1

MidoriKami:
Fix some filter windows not working correctly


MidoriKami:
Seems the libs compare is still including `.git` in the url, and thus breaking the compare

reiichi001 (approved):

PR #5051, opened 2024-11-23
EzWondrousTails v3.2.0.3

MidoriKami:
Actually fixes incorrect scaling on border nodes in the book.


MidoriKami:
Went the wrong way on the scaling with the last update... oops.

philpax:
Man, your plugins are cursed. I hope goat can get to the bottom of this at some point.

MidoriKami:
It's specifically using dalamud.NET.sdk inside library projects that's not working correctly in plogon. It used to be fine, but something changed with dalamudpackager (I assume) that's making it not play well anymore.

Goat suggested in another post, that releasing a version of dalamud.NET.sdk that doesn't include dalamud packager, would make it suitable for use in library projects like KamiToolKit and KamiLib.

reiichi001 (approved):

PR #5058, opened 2024-11-24
SortaKinda v2.1.0.2

MidoriKami:
Fix filter selection window scaling logic


MidoriKami:
Related note, it seems Dalamud.ImClip.Clipped draw behaves differently in 7.1 than it did in 7.0? It seems to be applying the scale internally somehow as removing my `* ImGuiHelpers.GlobalScale` is making things behave as expected.

reiichi001 (approved):

PR #5109, opened 2024-11-28
ResLogger2 0.0.2.8 -> 0.0.2.9

perchbirdd:


perchbirdd:
Note that almost all new code is unreachable - the initial PapHandler is commented out until I am able to determine a solution regarding incompatibility with an extremely common third party plugin.

reiichi001 (approved):

PR #5123, opened 2024-11-29
EzWondrousTails v3.2.0.4

MidoriKami:
Fix current duty indicator


MidoriKami:
Note: All 4 plugins that use KamiToolKit were updated to the same version of KamiToolKit

reiichi001 (approved):

PR #5134, opened 2024-11-30
Magitek Clicker [Testing]

ereti:
it's a plugin for playing clicker sfx in response to chat messages.

it's pretty small but i figured it was worth submitting given that i am sure others will get use out of it


reiichi001:
bleatbot, approve

ereti:
oh, i don't actually think i'm even *using* that

reiichi001:
Hi there! I've added this plugin to our review queue. There are a few things that I noticed which will need to be addressed:
1. Plugin icon appears to not be in the correct format. 
2. Your repo URL is set to something else in <https://github.com/ereti/MagitekClicker/blob/master/MagitekClicker/MagitekClicker.csproj#L8>
3. To use NAudio, you'll either need to implement it without windows forms being required (try using these nuget packages directly instead of full fat NAudio: `NAudio.Core`, `NAudio.Vorbis`, `NAudio.Wasapi`, `NAudio.WinMM`) or add `<EnableWindowsTargeting>true</EnableWindowsTargeting>` to your csproj to use them. Here's a recent example of when someone switched https://github.com/MKhayle/EldenRingDalamud/commit/a065ed68492df1753e98bca47713db7060503c11#diff-39fd6c532af4b6c2bfa37c67bf64ad416cafb86b03e0d8d2a08c7dca3f231282R58

I also see you're using `Dalamud.Plugin.Bootstrap.targets`. This is being deprecated in favor of a custom .NET SDK solution. To future-proof things as we may eventually remove the targets file, you can swap these lines out. 
```
<Project Sdk="Microsoft.NET.Sdk">
  <Import Project="Dalamud.Plugin.Bootstrap.targets" />
  ```
  to
  ```
<Project Sdk="Dalamud.NET.Sdk/11.0.0">
```
  The major version will always match Dalamud, but it will get minor updates every now and then.

ereti:
got it - i'll fix 2 and 3 now.
i'm not sure what you mean about the icon though? it's a 128x128 png, and the only guidance i can see on it is that it should be a png between 512x512 and 64x64

reiichi001:
You'll want to remove `NAudio	2.2.1 from your nuget packages when swapping to the more cross-platofrm friendly sets. It may have a few minor namespace changes.

ereti:
to be clear, am i able to keep NAudio if adding `<EnableWindowsTargeting>true</EnableWindowsTargeting>`? that's what i tried to do above, and i don't understand the error i'm getting - i've tried switching to using the packages you recommended but i'm getting one long continuous headache trying to switch from AudioFileReader (which only exists in NAudio proper) to VorbisWaveReader

philpax:
@rekyuu I know you've worked on NAudio-using plugins before; do you have any particular insight into what's happening here?

rekyuu:
The way I got it to build on Linux was to add the `TargetFramework` property and set it to `net8.0` instead of `net8.0-windows` as specified in the Dalamud targets. Maybe try doing that and removing the `EnableWindowsTargeting` thing? I can't see the build log but if you're getting failures to do WinForms dependencies that should probably fix it.

philpax:
bleatbot, rebuild

philpax:
Hi there! Have you had a chance to try out rekyuu's suggestions? Do let us know - otherwise, I'll have to close this in a week or two.

PR #5201, opened 2024-12-08
ChatProximity 1.3.1

iv230:
## 1.3.1
- Fixed targeting/targeted color switch

_Seriously, this one a pretty dumb issue_


iv230:
After build bot said previous version was 1.2.0.0 but it's actually 1.3.0.0, so the diff is incorrect. However it points to the good commit so the correct version would be merged.

philpax (approved):

PR #5207, opened 2024-12-09
[Testing] MagitekStratagem v0.11 - Builtin server

meoiswa:
Testing a followup idea to separating the plugin and tracker integration code:
This experimental new version of the plugin includes an entrypoint with which the plugin can be executed as the server.
This allows a separation of processes between the plugin + client, and native integrations + server, without requiring users download an external dependency, and mantaining all code compiled by the pipelines (hopefully this works!)
Unfortunately this also means the plugin itself is much bigger, as it has to bring along the required dependencies to run the AspNet server. Hopefully this won't be a major issue.


philpax:
Have you been able to investigate this / figure out a way forward?

meoiswa:
As far as I understand, there are .net packages that need to be whitelisted in order for the solution to build.

philpax:
Great, I'll hand it over to @goaaats to figure out then

philpax:
https://github.com/meoiswa/MagitekStratagem/compare/f9ea4199708206ab56a7a52705d33d42e59fa729..5829bb09edacb82bafe7014eeee8bc5f053769a9

philpax (approved):

PR #5235, opened 2024-12-15
ItemVendorLocation v2.9.0.1

electr0sheep:
- Fix problem when not every vendor for an item has a shop
- Fix problem where menu selection matching wasn't specific enough (multiple things highlighted)


electr0sheep:
I found another bug before this PR was merged, so I just pushed it up. Hopefully that's fine.

PR #5247, opened 2024-12-17
ItemSearch Plugin v1.10.1.0

MKhayle:
- Added a By Patch filter option (thanks @Kouzukii for the data I absolutely stole without pity from Which Patch Was That)
- Added right clicking on most options to revert them to their default state
- Fixed the Try On issue where it didn't apply the second dye when using the up/down arrow keys
- Fixed the size of the selected item box to make it the same size as unselected to avoid misclicking in the list by @smrq
- Fixed the size of the job selection area in the Equip as filter to avoid clipping because @SheepGoMeh complained about it


MKhayle:
Sorry for the size file, it's actually just raw data imported through WhatPatchWasThat.

reiichi001 (dismissed):

reiichi001 (approved):

PR #5335, opened 2025-01-07
QuietDen - Moving plugin from testing to stable

xorelcodes:
Moving over from testing, working as intended


reiichi001:
Since you're on `Dalamud.Net.Sdk`, you don't need the `Dalamud.Plugin.Bootstrap.targets` file anymore. The SDK has replaced that functionality. But that can be handled next plugin update.

reiichi001 (approved):

PR #5360, opened 2025-01-17
[Testing] Phlegma Dilemma

QLEDHDTV:
Phlegma Dilemma is an action rangefinder, visualizing the skill's range, radius, it's shape and the potential targets that can be hit if the action is AoE. Also provides a ruler to visualize the distance between the player and the NPC/object.


philpax:
bleatbot, approve

reiichi001:
The owner field will need to be adjusted to your github username, `QLEDHDTV`.

philpax:
Hi, this plugin is now in the review queue. Would you be able to include a screenshot of what it looks like in use?

philpax:
Apologies for the extremely late reply. After much internal deliberation, we have decided that this plugin runs the risk of upsetting the balance of combat, and as a result, we cannot allow it on the main plugin repository. We're very sorry about both the time it took to reach this decision, as well as the negative result, and we'll revise our combat rules in future to better clarify this.

PR #5368, opened 2025-01-21
Collections v1.0.2.1 - Update for API 11

Seventhxiv:


Seventhxiv:
Why is it detecting API 10 ? Did I miss anything ?

reiichi001:
Diff: https://github.com/Seventhxiv/Collections/compare/0292519f4ca7a619b9b99df0b90fa122e6e92c90..c6a62ea3bc4298640689925de07be48014204536

DalamudPackager is automatically imported when using Dalmaud.Net.Sdk. Adding it to the project will override that. (Same with TargetFramework)

But specifically, it's because of https://github.com/Seventhxiv/Collections/blob/master/Collections/Collections.json#L19 . (Let DalamudPackager handle that field for you. If you manually specify it, it will override DP)

reiichi001 (approved):

PR #5382, opened 2025-01-24
Dalamud.RichPresence 2.0.6.5

reiichi001:
nofranz


philpax (approved):
we should probably have some kind of policy around binaries

reiichi001:
iirc, Blooym tried to have it pulled programmatically, but it wasn't working reliably. Alternatively, we could/should package the Wine bridge with xlcore (this how XOM handles this dependency for macOS users). Then other plugins could make use of it too.

PR #5390, opened 2025-01-26
Honorific 1.5.1

Caraxi:
- Added colour previews in the character config window.
  - Can be disabled in main plugin config.
- Added location based condition.
  - Housing zones can be further filtered by Ward, Plot and Room


Caraxi:
Not in the primary changelog but I also added detection for invalid titles for mare.

philpax (approved):

PR #5445, opened 2025-02-12
[Testing] Tabnado 1.4.7

Paparogue:
Tabnado is a plugin that improves Final Fantasy XIV’s default tab targeting by offering a camera based approach inspired by the targeting style in World of Warcraft. Instead of using automated logic or prioritizing targets based on conditions like low HP, highest priority, or best AoE target, Tabnado relies on manual thinking and input, allowing you to switch targets smoothly and intuitively without giving any unfair advantage. Think of it as an alternative type 3 targeting method that fits naturally into normal gameplay, ensuring a fair experience for all players.


reiichi001:
bleatbot, approve

philpax:
Hi there,

Apologies for the late decision on this, but after discussing it internally, we believe that this plugin affects the flow of combat too significantly to fall under our combat plugin guidelines. We're sorry about this result, and we welcome you to submit another plugin!

PR #5464, opened 2025-02-23
Title Edit v3.0.4.8

RokasKil:
- Fixed a rare on load issue
- Added the option to hide vanilla and built in presets
- Updated the about section


RokasKil:
Also added an April Fools easter egg but that doesn't need to go into the full changelog :)

reiichi001 (approved):

PR #5480, opened 2025-03-01
CurrencyAlert v2.2.0.6

MidoriKami:
Fix border not rendering correctly


MidoriKami:
Most of the kamilib changes are old changes that have already been merged into my other plugins (like NoTankYou), there was only a 1-line change to change the background texture size from 32x32 to 64x64, because scaling is handled automatically now.

PR #5487, opened 2025-03-02
[Testing] AcquisitionDate 0.2.0.0

Glyceri:
[0.2.0.0]
- Fixes downloads not working.
- Build the system around parsers instead of parsing directly in web requests.
- Language based date formatting
- Now shows if a date is the earliest date and potetially not accurate or could be acquired before a given date.
- Downloads now parse the pages language instead of asuming the page language.
- Downloads should be more stable in general.

! The plugin is undergoing majour reworks which had to be pushed earlier than I expected.
! You should see no issues but there's a lot of ""framework"" and unused code in this update.

### DIFF
--- a/testing/live/AcquisitionDate/manifest.toml
+++ b/testing/live/AcquisitionDate/manifest.toml
@@ -1,10 +1,16 @@
 [plugin]
 repository = "https://github.com/Glyceri/AcquisitionDate.git"
-commit = "0d12448e1c4fdc691b06a711434f57d8878028f5"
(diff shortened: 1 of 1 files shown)


Glyceri:
This plugin was a mistake....

"Wouldn't it be cool to see the dates you got achievemets and what not"
Ye... NO! NO IVE CHANGED MY MIND I HATE IT NOW!!!!!!

reiichi001 (approved):

PR #5490, opened 2025-03-03
Absolute Roleplay [Testing]

serifas:
Updated the bio tab mainly to allow for more customization


reiichi001:
It's a bad practice to include the bin/obj folders generated from your machine's local builds. I would recommend removing those from your git repository first and then add those paths to a `.gitignore` file. GitHub provides one tailored to ignore common Visual Studio / C# artifacts here: <https://github.com/github/gitignore/blob/main/VisualStudio.gitignore>

~~As for the submodule error, I think you need to specify the branch you want to follow. Tryin using this:~~ I see this was fixed while I was typing.
```
[submodule "OtterGui"]
	path = OtterGui
	url = https://github.com/Ottermandias/OtterGui.git
	branch = main
```

serifas:
I will thank you

PR #5525, opened 2025-03-15
[Testing] Racingway v0.0.5.8

Abyeon:
- Fix routes tab using old export code
- Add timer overlay


Critical-Impact (approved):
LGTM

With your use of ImGui.SetWindowFontScale, it'll work however I would recommend building your own font at the size required and pushing the font

Abyeon:
Yeah I agree, I'll probably be doing that tomorrow (I got lazy). I don't know the process of adding a font to ImGui so I'll definitely be around in discord sooner or later asking about it.

PR #5535, opened 2025-03-16
Razer Reader Stable Release

mashirochan:
I kind of forgot about this plugin for a bit, oopsies.. I have been using it for months now with no issues and I haven't had any issues reported, so should be safe to put on stable (no changes have been made either, I don't think).


philpax (approved):
Good luck! May you get positive feedback from the hordes :)

PR #5536, opened 2025-03-18
New Plugin: Quest Share

nathanctech:
This new plugin allows players to "share" quest progress with each other through a simple interface, allowing friends to follow each other through quests they've otherwise completed. It will show the current progress of the quest and provide a marker to the next step's destination. When enabled, it will automatically track MSQ quests as they are accepted.

This uses an external server to handle requests which is also [open source](https://github.com/Era-FFXIV/QuestShare.Server).


Critical-Impact:
Just completed the code review, only one thing minor code thing

Should you be waiting on this? As far as I can see if you don't the API connection could still be shutting down after the plugin has finished unloading
https://github.com/Era-FFXIV/QuestShare.Plugin/blob/b3932d3388371b81e0e48203bf6024fe4408f26b/QuestShare.Plugin/Services/ApiService.cs#L79

In the grand scheme of things probably not a huge issue but just thought i'd flag it

nathanctech:
> Just completed the code review, only one thing minor code thing
> 
> Should you be waiting on this? As far as I can see if you don't the API connection could still be shutting down after the plugin has finished unloading https://github.com/Era-FFXIV/QuestShare.Plugin/blob/b3932d3388371b81e0e48203bf6024fe4408f26b/QuestShare.Plugin/Services/ApiService.cs#L79
> 
> In the grand scheme of things probably not a huge issue but just thought i'd flag it

Ah, good point. I'll make a quick fix tomorrow so it waits. I haven't run into issues when loading/unloading a bunch but you never know with async things.

nathanctech:
bleatbot, rebuild

reiichi001:
bleatbot, rebuild

Critical-Impact (approved):

PR #5549, opened 2025-03-21
Pet Nicknames v2.3.0.14

Glyceri:
nofranz


philpax (approved):
I did notice that I approved the plugin update *after* merging the denylist, but I figured it wouldn't be an issue. Whoops.

PR #5587, opened 2025-03-26
[1.1.3] ShopItemRevealer

nathanctech:
API12 update, no functional changes.


nathanctech:
Committed to test on accident, will resubmit to stable.

PR #5709, opened 2025-03-27
Sightseeingway: First for "not on main thread" exception

LeonAquitaine:
Fix for "not on main thread" exception


philpax (approved):
second

PR #5770, opened 2025-03-28
SimpleMarketBoard 1.9.5.1

Elypha:
Hi

just an update to api 12


Critical-Impact:
Hi your friendly PAC member here

I would advise against caching the LocalPlayer, you should access it and check it's validity each time you use it. 
You appear to be using it to check their home world/current world. I would poll within Framework Update and cache those world ids directly.

Line in question
https://github.com/Elypha/SimpleMarketBoard/blob/503b60f932357891c3aefe12e95196b3421db080/SimpleMarketBoard/Plugin.cs#L274

reiichi001 (approved):

PR #5887, opened 2025-04-01
DailyDuty v5.2.2.2

MidoriKami:
Fix being unable to select raid duties in Raids Normal module


MidoriKami:
I intentionally revert the changes made in the last version. The cause of the issue wasn't related to that code, so it didn't need to be included.

reiichi001 (approved):

PR #5928, opened 2025-04-06
Tf2CriticalHitsPlugin 3.7.0.0

jkchen2:
Update for API 12


Critical-Impact:
This seems fine and I'll merge it, though quick question, is there a reason you are running on a fork of KamiLib?
As far as I can your fork is a fork of Berna-L's fork and that's a fork of the original KamiLib which is still getting updates so if you aren't wanting custom functionality it might be easier to switch the submodule to that one.

Will probably get reviewed faster as well as the submodule then has a higher chance of having already been reviewed :)

Critical-Impact (approved):

PR #5933, opened 2025-04-06
PennyPincher 1.8.0.5

tesu:
* stop using network packet processing because it's bugged (thanks @nebel)
* minor bugfix to stop copy garbage value upon `Please wait and try your search again` error


Critical-Impact:
I do think this is really deucalion's problem to fix but thank you for fixing it up :)

PR #5962, opened 2025-04-09
Wave Wrangler Stable Release

mashirochan:
It's such a simple plugin, I don't think it's possible for anything to go wrong with it..

Should be ready for stable!


philpax (approved):
Waves: wrangled

PR #5977, opened 2025-04-12
[Testing] [New Plugin] PartySortPlus v1.0.0.0

Rietty:
This plugin allows for sorting your party list given a predefined preset of job ordering and allowing to control for player's job, party members job and instance you are in. It operates on territory change and can also be run manually via `/psp`


philpax:

Rietty:
Hi!

This plugins sole intention is to simply allow for some more flexibility
when it comes to the base game’s existing psort functionality, as such I
limited the design in such a way that people can do the following:

They can have many different orderings for job priorities, aka the
“presets” in the plugin.

Rule wise there can be as many rules as they want however the only things
that can be checked for are:

- Instance or Territory you are in.
- Job(s) that the plugin user is playing
- Jobs present in the party.

Each rule can have only one associated preset.

Rules run in a top down fashion and stop after the first rule is run on a
territory change.

The plugin runs automatically when user loads into a new area, or when
manually run with /psp like if they did /psort.

Running the command again does not sort it differently unless different
rules are enabled/disabled.

There is no plans now or in the future to make it automatic based off
combat conditions since I would like to maintain it as a main-line repo.

philpax:
Great, thanks! That's satisfactory for us - we'll keep it going 🙂

philpax:
bleatbot, rebuild

philpax (approved):
Thanks for responding so quickly to feedback, and congratulations on another plugin! 🙂

PR #5992, opened 2025-04-13
[Testing] Render Adjust alpha

Glorou:
Plugin allows users to dynamically adjust the in-game object limit or to set it to a custom value.


philpax:

Critical-Impact:
Hi, PAC member here, I have reviewed your plugin and it looks good barring these minor things:
I'm marking it as approved but feel free to push up a commit before or after

I'd recommend making sure that log calls like this are set to verbose as they may fill up the log quickly
https://github.com/Glorou/RenderAdjust/blob/8c2b6101238b656c1e529938d11c8e38676dd2d0/RenderAdjust/Plugin.cs#L124

You could store this instead of scanning for it each time you wish to override(as it's part of the game itself it's memory location won't change once found.). I would also wrap where you ScanText in a try/catch as if when a patch hits the address of this changes, it will throw an exception
https://github.com/Glorou/RenderAdjust/blob/8c2b6101238b656c1e529938d11c8e38676dd2d0/RenderAdjust/Plugin.cs#L175

Glorou:
I've updated with your suggestions, and patched a bug I found as well, thanks @Critical-Impact

philpax:
I've sent this off for testing now and hope to get back to you soon.

Another PAC member raised a potential concern: the setting shouldn't go below SE's minimum, as it's possible that you may end up hiding vital actors (i.e. the boss) due to the code assuming that they would always be present and thus not applying any prioritisation. I'm not sure if this is an issue as currently implemented, but we'd figure we'd call it out as a potential risk.

philpax:
@Abyeon tested out the plugin and reported this:
> Loads/unloads fine, but doesnt actually seem to do what it says. Tried playing with the values a bit, but never actually saw the character limit change. Might try it out in a hunt train or something though.
> Also, doesnt have a main UI callback

philpax:
For reference, Glorou and Abyeon discussed the issue and came to an agreement over the current behaviour and what the behaviour should perhaps be. As this has been tested, I'm happy to merge it as-is, but I'll let you decide if you want to update it or not first.

Glorou:
Hey Phil, 

I've updated with the suggestions to be more robust at locking out users from using the custom setting while in possibly game breaking scenarios. 

Thanks!

philpax (approved):

philpax:
Congratulations on your first original plugin being added to the main Dalamud plugin repository!

It looks like we didn't give you access to the private developer channels for plugin feedback and others when you adopted your last plugin; my bad! If you're interested, let me know and I'll have the roles granted for you.

Glorou:
That would be great of you don't mind!

PR #6007, opened 2025-04-16
[Testing] [New Plugin] PvP Announcer

spide-r:
This plugin will cause Metem's voice lines to play mid-match upon specific conditions being met. This should not provide any material advantage in PvP as this reacts to actions after they have already happened.

Rough Outline of How this plugin works:
I Hook Into various actor functions (currently just actor control & action effect) and send all of these to a broker, which emits an event to various event listeners, which determine if a voice line should be spoken.

Most Event Listeners are focused on actions that are undertaken by the plugin user. However a few event listeners trigger a voice line when an enemy performs an action _and_ the user is hit by their attack. If the attack does not hit the user, nothing is triggered.


spide-r:
I've removed the ability for the user to enable tracking of their teammates. Mostly due to complexity reasons. Figured i would mention if that was a potential issue on the PAC side. I've also gotten most of the necessary events tracked and registered. Apologies for the consecutive edits.

Critical-Impact:
Hi, PAC member here, just reviewing the code, nothing I can see that'll stop an approval that said

https://github.com/spide-r/PvPAnnouncer/blob/8af1824d6c2091c4d6f3ee5d12fd15990cec27e7/PvPAnnouncer/Impl/PvPEventHooksPublisher.cs#L28
I'd recommend putting your code in a try catch, any code inside a detour should have it IMO

spide-r:
Understood. Thank you for the catch. I'll implement soon!

spide-r:
Bug Fixes, Minor Additions, Removal and Cleanup ^

spide-r:
It's ready now I promise!

Critical-Impact:
Hi, one of our testers ran into a minor issue, the hooks appear to be leaking when the plugin is unloaded.
Specifically I don't think this class `https://github.com/spide-r/PvPAnnouncer/blob/master/PvPAnnouncer/Impl/PvPEventHooksPublisher.cs` is being disposed of anywhere.

If you can take care of that then that should get it to the point where we can this merged

philpax (approved):
coming in to steal this merge from Critical who did all of the hard work

PR #6045, opened 2025-04-22
[Testing] HUDManager 2.5.22.0

nebel:
- Add support for new HUD elements


reiichi001:
You put your HUDManager commit hash in the PartyIcons folder/manifest.

nebel:
Yeah that seems like something which would break the build. Sorry.

reiichi001 (approved):

PR #6046, opened 2025-04-22
[Testing] Moonbase Omega 1.0.0.0

NotNite:
Adds Moonbase Alpha TTS when in Cosmic Exploration. aeiou!


NotNite:
Some technical notes for PAC:

- DECtalk is interacted with via `LibraryImport`s. This is technically unsafe code.
- This intentionally only works in Cosmic Exploration (ID 1237, path `ffxiv/cos_c1/hou/c1w1/level/c1w1`) for comedic effect. The TTS part can be tested by using the `/moonbaseomega debug` command (e.g. `/moonbaseomega debug john madden`). I would go into Cosmic Exploration and test the chat channels manually, though (maybe Echo?).
- I create multiple DECtalk instances since DECtalk can only do one string at a time. This limit is hardcoded to 5 (see SpeechManager.cs), but I might make it configurable in the future.
- DECtalk is not distributed in my own repository, instead I download it automatically off of the releases at https://github.com/dectalk/dectalk. I hardcode the release and SHA256 check the zip file (see DownloadDecTalk in Plugin.cs). These get stored in the plugin config directory, and it's a very small .zip, so the download runs automatically on startup if the files are missing (and pops up a notification if something bad happens).
- DECtalk is abandonware, I'm not really sure what's going on with the licensing? I've made care to not redistribute it myself, at least. I don't think the company that owns DECtalk even exists anymore.
- The TTS bit and a testing CLI app is also present in the repository. The TTS bit is a library that the plugin uses.

NotNite:
Note to PAC: I have found a few minor issues with the plugin, but it doesn't impact the plugin's functionality or cause any issues with the game (tmk). If these are blockers for a merge, let me know, but I won't be able to fix them until next week: https://github.com/NotNite/MoonbaseOmega/issues

Critical-Impact (approved):
aeiou!

PR #6068, opened 2025-04-27
Clean Hunt Chat v1.2

Glorou:
I've added the ability to let users enable this setting permanently. It's on the PermSetting branch (I'm not sure how you're supposed to branch for the sake of review so please let me know if I did this wrong).


reiichi001:
Commit hashes take care of source branch. Please make sure to bump the version, as that is a hard requirement for all plugin merges.

PR #6071, opened 2025-04-28
NextUIPlugin stable version 11.1.1.0

pwilkowski:
Updated for new api.
Fleck has been pulled directly into code as nuget package was outdated but i did find a fork on github that as maintained and worked.


reiichi001:
This might've been a case where it would have been preferable to submodule the other repo instead of cloning it directly into your source code, which would have simplified the review scope just a tad.

reiichi001 (approved):

PR #6089, opened 2025-05-02
[Testing] Moonbase Omega 1.1.0.0

NotNite:
- Added a force stop button to make the TTS shut up
- Added "Auto Delete Log File" option to clean up DECtalk's mess (disabled by default)
- Added configurable instance limit
- Added a warning for when DECtalk fails to download
- Fixed map and item links being read out loud
- Fixed config being written too many times when adjusting settings
- Fixed a CTD when adjusting the TTS volume in the middle of a sentence

### DIFF
--- a/testing/live/MoonbaseOmega/manifest.toml
+++ b/testing/live/MoonbaseOmega/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/NotNite/MoonbaseOmega.git"
-commit = "d0baf7c7355f8223e34dca30bba21e6bc01ef5d9"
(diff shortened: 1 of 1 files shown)


NotNite:
Note to PAC: you heard that right, there's a CTD! It only happens when adjusting the volume during a spoken sentence, and it only happens *sometimes*, so I don't think it needs to be banned immediately (I'd rather not leave people without a working version of my plugin). Please ban 1.0.0.0 once this is merged, though (or I can do it if you need).

philpax (approved):
john madden

Please make a PR to DalamudAssets for the ban and tag me, I'll merge it

PR #6113, opened 2025-05-06
SettingsEnhanced 0.1.1.0

Blooym:
- Settings names are now localized based on your game language.
- Some instance content types are now supported (dungeons, trials, raids, alliance raids)
- UI Rendering performance has been massively improved
- Changed all UI code to use ImRaii


philpax (approved):
Your choice of timing has been noted for next quarter's performance reviews.

PR #6155, opened 2025-05-15
Goodfriend v3.5.0.0

Blooym:
Updated for Dawntrail with a first-pass cleanup on top.

## Developers note on using this plugin after 7.0
As of Patch 7.0, some of the data that GoodFriend uses to identify players during login/logout events is now visible to anyone nearby your character in-game (not just your friends or party members). GoodFriend itself will still only show notifications for people you've added to your in-game friends list - however, please use this plugin only if you're comfortable with your login/logout activity being theoretically visible to nearby players (note: GoodFriend never sends data for users who don't have the plugin installed).

If you ever want to stop sending login/logout events, simply disable or uninstall the plugin. The server does not store any data about you.

**For added privacy, you can now set a 'group key' in the plugin settings**. This allows events to be shared only with people using the same key (for example, everyone in your raid group). When a group key is active, your events are never visible to anyone outside your group no matter what.

### DIFF
--- a/stable/GoodFriend/manifest.toml
+++ b/stable/GoodFriend/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Blooym/GoodFriend.git"
-commit = "5f9261585043464bf82f0c8caada7587104ddfce"
+commit = "26dd9df738d036530d080ed4d335789082f94291"
 owners = [
     "Blooym",
 ]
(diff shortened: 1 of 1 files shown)


Blooym:
Discussed with PAC before submitting for 7.0

philpax (approved):

philpax:
rude of you to add a commit while I'm reviewing it 😤

Blooym:
sorry!!!

PR #6163, opened 2025-05-16
GoodFriend v3.6.0.0

Blooym:
Removed the [GoodFriend] tag from some messages. Also fixed the current friends online module not outputting due to an exception.


philpax (approved):
the rocket has crashed, F

PR #6220, opened 2025-05-29
GoodFriend v3.7.0.0

Blooym:
nofranz

Inbound events are now sent using message pack, putting this in until I rewrite to help save some bandwidth


Blooym:
Need to setup infra changes to support this actually, going to hold off for a bit

philpax (approved):
*approves while sobbing*

PR #6229, opened 2025-05-30
CurrencyAlert v2.2.1.3

MidoriKami:
New style configuration system

**Your old style settings will be reset**


philpax (approved):
As always, I hope you know what you're doing by resetting users' configurations and raising their ire in the process

PR #6344, opened 2025-06-29
[Testing] Photobooth initial release

iri-impulse:
Photobooth provides a companion window for the portrait editor with improved/advanced controls.

First time making one of these; this little side project got more than a little bit out of hand. Let me know if there's anything I ought to do to make it more convenient to review.


Critical-Impact:
Hi, it appears that your PR is incorrect as it points to https://github.com/iri-impulse/Photobooth and not https://github.com/iri-impulse/PhotoboothPlugin
You'll either need to change the repo name or PR

iri-impulse:
@Critical-Impact whoops sorry, should be fixed now. I bumped the commit because one of my tester volunteers caught something especially confusing but don't anticipate having to touch that again until after I've given reviewers a chance to get through it.

Critical-Impact:
bleatbot, approve

philpax:
bleatbot, approve

Critical-Impact:
Hi, I have completed the code review for this, just after some clarification.
From what I can see it does have safety checks in place to match what the game allows. How accurate is this and do you think it'd be possible for the plugin to cause invalid portrait data to be sent to the server?

iri-impulse:
Luckily the client is very thorough about checking the camera stuff -- it has to be, you can get the camera to all kinds of weird places with the default UI already. So _those_ checks if I've messed up the math on will generally just mean the save button is disabled. I'd want to advise testers not to attempt to save the portrait if it looks obviously bogus, but between the save button check and the RE'd bounds here I'm not very worried; I got the camera into a lot of funky positions early on and they were all reliably not saveable.

The lighting stuff is straightforward to check because they're just sliders in the default UI. I actually got both of these wrong at some point during development so I (unfortunately) know how the server behaves when they're bogus.
- Lighting colors: The minimum brightness is 20; apart from that all unsigned bytes are legal values. _Neither the client nor the server_ actually checks the minimum brightness and the portrait saves and displays fine, but obviously I'm still forbidding edginess values below 20.
- Lighting direction: The two angles here are saved as shorts but the valid range is only [-180, +180]. This also needs (and has) explicit enforcement, because angles outside that range look totally normal on the client (it just wraps around) but the server rejects them.

These are somewhat less straightforward to verify, but seem to be robustly enforced by the game. Even if you try to put them out of range, they get reset to a valid value each frame.
- Head/eye direction: I determined the bounds experimentally (by making the character look at the camera). I didn't actually check if some races have more or less flexible necks, because the good news here is the game already clamps the angles: if you set the plugin's bounds to something wider, you still can't drag the direction past what the model permits.
- Animation progress: the game engine makes it impossible to go past the end of the animation here, which is good because the way I'm estimating the max progress isn't exact. Every value between 0 and the end is theoretically accessible in the default UI, though some require an unsteady frame rate -- if you're locked at a perfect 60fps some progress values will fall into the gaps between frames every loop.

I haven't done anything with the unlockable stuff (poses, backgrounds, accents, frames) so there's nothing to worry about there. If I do add that later on I'll definitely have to be careful about only letting you use things you've unlocked.

I think that covers everything. The server interaction here happens when you hit save, since even though non-adventurer-plate portraits aren't stored on the server they need to have a checksum computed by the server and saved alongside them.

philpax:
Thanks for the thorough review! I'll send this off to testing now 🙂

philpax:
bleatbot, rebuild

philpax:

iri-impulse:
@philpax display name there is `iri` as well -- don't really want to leave my discord handle googleable, I can send a message there if that's not specific enough

iri-impulse:
(Discord username's sha1 hash is `3e3571a29f3beb38b6190c411e9e224584d79c36`)

PR #6383, opened 2025-07-10
add new plugin AetherBreakout

rail2025:
Im not sure why all these old commits are being included with the pr. i probably did something wrong with github desktop. sorry, let me know if i need to do some other way, not very familiar with that.

### DIFF
--- a/stable/ChillFrames/manifest.toml
+++ b/stable/ChillFrames/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/ChillFrames.git"
-commit = "0a7fc07f27ca5eeee6e1d01cc7bee97493f9409f"
+commit = "0bf9ea1c276243428cd56e8b5bbd125bd2fc8181"
 owners = ["MidoriKami"]
 project_path = "ChillFrames"


nathanctech:
Error from build:   `[09:11:46 INF] [CONTAINER] /usr/share/dotnet/sdk/9.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(425,5): error NETSDK1073: The FrameworkReference 'Microsoft.WindowsDesktop.App.WindowsForms' was not recognized [/work/repo/AetherBreakout.csproj]`

Looks like you're including NAudio which is Windows only. You might have to specifically include your target framework to make Plogon happy in your csproj:

`<TargetFramework>net9.0</TargetFramework>`

I believe there's a couple other plugins out there that also use NAudio, but don't exactly remember how to make it build properly. Maybe it doesn't like the NAudio.WinForms reference?

As for your Git, you'll need to "sync" your fork of the plugin repository before adding commits or it will get messy :) happens to everyone. I highly recommend using a branch on your fork that only contains commits for the plugin you're working on to prevent messes too.

PR #6385, opened 2025-07-10
Housing Herald Testing Release

mashirochan:
This plugin might be a bit of a hassle to test... sorry!

How it works:
1. Player bids on house
2. Bid information is stored in plugin config
3. Leading up to Results Phase, player is reminded upon login which house they bid for and when Results Phase starts
4. Once Results Phase comes around, player is reminded upon login which house they bid on and how long they have until Results Phase ends
5. If Lifestream (custom repository plugin) is detected as loaded, then the player will get the option to teleport to the bid plot
6. Once the player checks their bid results, the notifications should no longer occur upon login
7. Once the Results Phase ends (new Entry Phase starts), the player's bid is cleared out from the config

At any time before or during the Results Phase, the player can use `/pherald` to open up the plugin UI to view their bid information.


philpax:
bleatbot, rebuild

Critical-Impact (approved):

PR #6393, opened 2025-07-13
AetherDraw update v2.2.0.0

rail2025:
added limit cut dots, raidplan.io import


rail2025:
I left the debug logs on the texturemanager and translator because it's helpful to identify new/unknown shapes/icons/emojis when loading new raidplan imports, and therefore easy to add to the dictionary. shouldn't affect end users performance and will speed up making imports look better.

PR #6394, opened 2025-07-13
Umbra 3.0.10

haroldiedema:
# Umbra 3.0.10

> Moving from Testing to Stable.

## Fixes & Improvements

- Fixed Auxiliary bars not respecting the configured Toolbar Height setting. 
- Removed the option to turn off Threaded Style Computation. It is now on by default.


haroldiedema:
Here is the actual diff from the last version that was in testing: https://github.com/una-xiv/umbra/compare/116033c8716bb2e7cf137c979130fcd4c153a760..603fa46ba51e21e4117172fa15ef2d9c1563fe59

PR #6427, opened 2025-07-25
[Testing] Portrait Helper v1.0.4

Haselnussbomber:
**Portrait Helper** allows you to save and load portrait presets, edit camera and pose settings in detail.

The plugin will automatically display a small bar with several options at the top of the portrait editor.

**Features:**

- **Preset Browser**  
  Save your favorite portraits as reusable presets. Simply double-click to load one!
    - Preset previews are saved as png files to the plugins configuration folder and contain the preset string as Exif metadata (in the UserComment field).
- **Clipboard Import/Export**  
  Share your portraits easily with base64-encoded strings.
  - If you want more control over whats being imported, use the Advanced Import Mode to selectively apply only the parts of a preset you care about.
- **Advanced Edit Mode**  
  Fine-tune your portrait with full control over
    - Camera yaw, pitch, distance, position (X/Y)
    - Zoom and rotation
    - Head and eye direction
    - Animation timestamp
- **Alignment Tool**  
  Displays on-screen guide lines to help with portrait composition and symmetry.
- **Reset Button**  
  Quickly revert any changes made since opening the Portrait Editor.


philpax:
bleatbot, rebuild

Haselnussbomber:
Even though it was short lived, I’ve decided to close this PR.

This plugin has a some issues that need to be worked on, mainly:

- The animation timestamp slider reloads the animation when scrolling backwards, and when scrolling forwards it doesn't update attachments, like weapons. This needs more reverse-engineering power than I can bring up.
- I don't like this big disclaimer about settings that might or might not apply and that you have to re-open the window if the game didn't like them.
- Presets are saved with a thumbnail of the current character. When you load a preset on an alt character, the result might look completely different from the thumbnail, depending on pose and race.
- There is no way to regenerate thumbnails, for example after taking a fantasia, without copying every preset string one by one, and importing them again.
- Adventurer Pass presets would also be neat to have.
(And probably more that I can't think of right now.)

So, I don't want to release it in the state it's in, and honestly, I have to admit that my motivation to work on this is at absolute zero  - for a long while now. The thought of having to deal with it again (ehm, I mean, to maintain it) once its in peoples hands is a mental toll right now which I rather want to avoid.

I'm very sorry for the trouble I've caused, especially for the code review. 😔 🙏

philpax:
Sorry to hear it, but understandable - take care of yourself!

PR #6431, opened 2025-07-25
NoTankYou v7.2.2.3

MidoriKami:
New rendering system for displaying warnings both in the PartyList and with the Warning Banners.

It's now super fancy!


MidoriKami:
Straight move from testing to stable. No changes, copied commit hash over.

reiichi001 (approved):

PR #6441, opened 2025-07-29
AetherPool v0.4.0.0 initial submission

rail2025:


philpax:
We've confirmed that this is rules-compliant and will be moving it through the rest of the pipeline, but we noticed that "Aetherpool" is already the name of something in the game: https://ffxiv.consolegameswiki.com/wiki/Palace_of_the_Dead#Aetherpool_Gear

You're not required to, but we suggest renaming the plugin to avoid confusion, especially for searchability.

Critical-Impact (approved):

PR #6452, opened 2025-07-31
Absolute Roleplay [Stable] [Testing-Live]

serifas:
sorry, I know my last one just got reviewed, I had a small issue with the my text wrap system not displaying correctly.
Also added another flag for the trees. that's about it.


reiichi001:
Frequent updates are fine! (And usually a bit easier to review). ...as long as it's not like, several updates a day, barring special exceptions like a release that crashes people.

PR #6515, opened 2025-08-07
PlayerTrack 3.5.2.0

Infiziert90:
- API 13

[thanks littlenightmare]
- improve the rules of the filtering world
- use luminaWorld.Name instead of luminaWorld.InternalName to sync with in-game text

[thanks nebel]
- Make local NameTextWrap nullable, and only set if non-null


philpax:
bleatbot, rebuild

reiichi001 (approved):

PR #6598, opened 2025-08-08
AetherFM – Migrate to source-based plugin (API 13 compliance)

SalvatoreDevelopment:
This PR updates AetherFM to support Dalamud API 13 and comply with AGPLv3 requirements.

- ✅ Full source code included under `testing/live/AetherFM/`
- ✅ Plugin is now built from source via `.csproj`
- ✅ Manifest updated with correct commit and `api_level = 13`
- 🔧 Tested locally with Dalamud.NET.Sdk 13.0.0 and .NET 9.0.3

This replaces the prior PR which used a precompiled binary.


reiichi001:
Please follow the submission guidelines like you did for prior releases of the plugin.

https://github.com/goatcorp/DalamudPluginsD17#publishing-your-plugin

You should only be providing a d17 toml file that points to a git repository and commit hash.

nathanctech:
Your plugin's csproj needs to have a different version number, or you need to specify it in the manifest. I see you intended on releasing it as 1.1.1 but did not update the csproj or specify it otherwise.

PR #6607, opened 2025-08-08
[Testing] CustomResolution v0.4.2.2

0x0ade:
- Update to API13
- Add even more self-verification to protect against the next game update.
- If this works fine, maybe it's time for 1.0.0.0


perchbirdd (approved):

reiichi001:
I'd highly recommend switching to using Dalamud.NET.Sdk so you don't need to specify DalamudPackage, Dalamud.Bindings, or API level in your json file

PR #6703, opened 2025-08-10
AdaptiveHud API13

Helios747:
API13


MidoriKami:
You need to regenerate your packages.lock file for it to specify api13 <https://github.com/Helios747/AdaptiveHudPlugin/blob/main/AdaptiveHud/packages.lock.json#L7>

You can use `dotnet restore` to regenerate this I think

PR #6708, opened 2025-08-10
DeathRecap 1.13.2.1

Kouzukii:
- Fix odd skill interactions when XivAlexander is in use


philpax (approved):
In future, we'd prefer that you not name custom repo plugins/external apps by name (we're trying to maintain clear separation), but I won't block this PR

PR #6716, opened 2025-08-10
DailyDuty v5.3.2.8

MidoriKami:
Guess whose back, back again, tell a friend, wondrous tails clovers in duty finder are back!

Honorable return to Wondrous Tails Instance notifications as-well!

With path 7.3 Square did us a favor and significantly simplified how they provide duty information for each wondrous tails task, so this feature is back stronger than ever!


MidoriKami:
no changes, straight move from testing -> stable, same hash

PR #6760, opened 2025-08-11
[Testing] BisBuddy v0.1.5.7

RajahOmen:
- Fix error being thrown when disabling plugin
 - Fix adding and removing gearsets not triggering inventory updates
 - Fix changing a gearsets highlight color changing the default color


RajahOmen:
found an issue with this - don't merge it yet please.

reiichi001 (approved):

reiichi001:
Oops. I had it up before the comment

reiichi001:
For a future case, I would change it to draft mode ASAP because that will prevent merges.

PR #6801, opened 2025-08-16
NoTankYou 7.2.2.9

MidoriKami:
Add action tooltip.

Can be disabled in Banner Settings.


MidoriKami:
No changes from testing -> stable, copied hash over.

PR #6808, opened 2025-08-17
CustomResolution v1.0.0.0

0x0ade:
Works on my machine, and apparently countless others now too. There's still ~~two~~ one and a half unresolved issues (one minor visual quirk on one PC only, and DLSS being quirky in vanilla XIV), but given that I see people shouting out cres already while not receiving any further bug reports myself, I'm biting the bullet to prevent people from hopping on testing when they shouldn't :')


0x0ade:
the fuck is that .vscode dir doing there

PR #6851, opened 2025-08-23
[Testing] VanillaPlus v0.1.1.2

MidoriKami:
Location Display - Fix not updating when changing map instances (such as instance 2 -> 3)

Lots of internal system changes - better performance, more reliable, blah, blah

DragableWindowDeadSpace - Removed Comptability disable, SimpleTweaks has updated, and no longer conflicts.
If you run into an issue, make sure both VanillaPlus and SimpleTweaks are updated

ListInventory - Rearranged UI a little, added "Reverse Sort Order" button

More internal changes - fixed scrolling elements actually being interactable even when not visible

Even more internal changes - fixed keybinds being triggered when typing into native game text inputs


MidoriKami:
PAC Note: this PR is only system changes and fixes, I will have another PR ready once this is merged that adds one new feature, details will be in the PR

PR #6894, opened 2025-08-28
Browsingway 1.6.7.0

Styr1x:
- Bump Chromium to 139.0.7258.139
- Improve renderer crash handling
- Fix CEF not loading when game runs as a privileged process (please don't..)


philpax (approved):
Honestly, I'd be inclined to just disable the plugin when the game is run as admin, but I don't have any strong feelings about it

PR #6903, opened 2025-08-30
Pet Nicknames v2.6.1.2

Glyceri:
+ Changed some IPC names. 
(If you are a plugin dev that uses the IPC, it should be extremely easy to fix).


Caraxi:
Why would you rename the API endpoint that is used to check which version of the API its using

Glyceri:
Because I'm me, its fixed.

Glyceri:
Oh my goooooooood

PR #6929, opened 2025-09-03
New plugin: FATE Helper

teechep:
added manifest and icon for FATEhelper


teechep:
sorry, forgot to add description here. FATE Helper will automatically level sync along with enabling tank stance, and automatically switch targets to forlorn maidens/the forlorn (optional). it also provides a menu with a list of all FATEs in the current area, and clicking the fate name will flag it and teleport you to the closest aetheryte

[repo](https://github.com/teechep/FATEhelper)

philpax:
bleatbot, approve

philpax:
I've added the plugin to our plugin review queue; we hope to get back to you soon.

The failure you're seeing is because you don't have a plugin manifest that describes your plugin; see the [SamplePlugin](https://github.com/goatcorp/SamplePlugin/blob/master/SamplePlugin/SamplePlugin.json)'s manifest. It should be named the same as your project and be filled in.

teechep:
sorry, added the manifest JSON now. thanks!

reiichi001:
Hi there!

Based on the description you've provided and looking at the code, I'm not sure we can allow this plugin on this main repository as it violates our [plugin guidelines](https://dalamud.dev/plugin-publishing/restrictions) for automating user actions sent to the server. 

Particularly:
- automatically pushing the level sync button on behalf of the user
- automatically applying tank stance on behalf of the user
- changing targets (this results in server packets sent to other players because you can see who/what another player is targeting) on behalf of the user.

We have a strict 1 user action via a plugin -> 1 server action requirement. While you can offer plugin UI buttons and then run those actions after the user has directly clicked/tapped/interacted with them, doing these kinds of actions without that would be classified as botting under our rules.

teechep:
the level sync only happens when a user targets a fate enemy, it's not automatic when they enter a fate. could selecting a target count as the user action to trigger that, or does it need to be more distinct like clicking a UI button? I'd be fine removing the tank stance and forlorn targeting if that would be ok.

similarly, when clicking the fate name it can currently flag the fate, open the map, and teleport; I assume only one of those 3 would be allowed. if I wanted 2 of those actions, would I need 2 distinct buttons, or could the same button do 2 actions when clicked sequentially (i.e. flag then teleport)?

sorry if I sound argumentative, just wanted to see what I need to change. thanks for your time and all your team's work on dalamud.

MidoriKami:
Hi, I'm **not** a member of the plugin approval committee, so I don't have any say in approval, but can provide some info with regards to server interactions, there are many things that we can do to the client that don't inform the server. 

Such as, placing flags, and opening the map. These don't send or request any data from the game servers. So these are allowed to be triggered by plugins, as there's no way for square to detect these actions.

Plugins need to operate in such a way that they are indistinguishable from a normal users network traffic, targeting an enemy and syncing to the fate in one frame is not something that a player can do normally, not even with a macro. And both of those actions send network data to squares servers.

Again, I am not a member that makes the call on matters like this, just wanted to give you some info as it seems your query has gone unanswered for a few days.

reiichi001:
(Kami, I appreciate that you took time to respond and helped clear up some of the rule explanations. It would have perhaps been better to check with one of us prior if you felt that a reply was delayed, to ensure there aren't/weren't any crossed wires. This kinda came off as  "I did this because PAC hasn't said anything yet" and has somewhat driven me to make a reply before gathering input from half the PAC members.)

I'll be quoting some remarks made in our private discussion areas, which I'm going to copy over and expand on if/as needed.

#### Point 1:
> The level syncing really comes down to whether or not we'd consider targeting an enemy an action and if we think it's permissible that plugins can take one action that's kind of unrelated and then perform another action because of that. That said I'd argue this violates our combat rules if they do go with what they've suggested. A button/ui element popping up would be fine but then it's not that different from what already happens.

I'm in agreement with other PAC members here. The player targeting an enemy is not a good premise for kicking off other server-bound actions. It does not directly imply the player intends to participate in a FATE. (What about gatherers or people flying over that may not want to be involved, but mis-click? Or someone who's trying to check the boss health or another player without intending to participate just yet?

While you could add some other button to prompt the user to join the fate / apply levelsync, there is already the [/levelsync](https://na.finalfantasyxiv.com/lodestone/playguide/db/text_command/9a2aa5ad272/) text command if the player wants to make a macro for a dedicated button. Drawing another one on screen through imgui or the like is fine, but at that point, they can also click the native one by the FATE information.

Adding a hotkey for levelsync would be fine, but you would need to ensure it doesn't cause other conflicts. (Or is user-definable)

#### Point 2:
> Tank stance/targeting = hard no unless they go some sort of ui with buttons as above

As mentioned in my prior reply, anything that automatically invokes an action would break our plugin guidelines. The same applies to automatically hard/focus-targeting another monster. This is making a decision on behalf of the player that also results in server data, as anyone can see who you have targeted.

#### Point 3:
> Flag fate, open map and teleport seems fine using a single button, other plugins do this

I don't have anything else to add here. Like Kami mentioned, opening the map and making a flag on it are client-side operations. A button that does that, and then kicks off a teleport would be fine, as there are already plugins doing similar, and the teleport is the only serverbound operation there.

teechep:
I've removed the level sync and tank stance/targeting. On the plus side there's now nothing running without the plugin window open so there's no chance of it affecting game performance lol. Thanks for the feedback y'all!

philpax:
Thanks for the update! We'll re-evaluate and get back to you 🙂

Critical-Impact:
Hi, just completing the code review for this and ran into these things:

1. https://github.com/teechep/FATEhelper/blob/b34463070eff95a2932d203f5826ed17a9cf033a/FATEhelper/FATEhelper.csproj#L6 - Please make sure you edit this and remove the references to SamplePlugin and remove any unused resources
2. https://github.com/teechep/FATEhelper/blob/b34463070eff95a2932d203f5826ed17a9cf033a/FATEhelper/Windows/ConfigWindow.cs#L15 - You've used `With a constant ID` as the ID for the window, while it's unlikely for another plugin to have made the same mistake but if they did your window and the other window would merge, please use a ID that would be unique to your plugin
3. Would recommend updating your README.md or removing it

teechep:
done done and done!

Infiziert90:
Did a quick look into your code and noticed that your MainWindow is using the same function in Draw and Update (which i have no clue why you would even do that), but also the if check you are doing doesn't prevent it from being run double each single frame and run everything, as the if is set to false, just to be set to true at the end again.

Also what is even more dangerous is that Update doesn't happen inside a BeginWindow frame in dalamud, as it isn't meant for any drawing at all, so you are drawing all your imgui content into a void which can lead to bad behaviour.

teechep:
> Did a quick look into your code and noticed that your MainWindow is using the same function in Draw and Update (which i have no clue why you would even do that), but also the if check you are doing doesn't prevent it from being run double each single frame and run everything, as the if is set to false, just to be set to true at the end again.
> 
> Also what is even more dangerous is that Update doesn't happen inside a BeginWindow frame in dalamud, as it isn't meant for any drawing at all, so you are drawing all your imgui content into a void which can lead to bad behaviour.

fixed. I thought Draw only fired once when the window was opened and then Update was for updating while open

PR #6958, opened 2025-09-07
Update manifest for CurrencySpender

Blackcatz1911:
# Changelog

### Added
- A new setting to automatically open Currency Spender together with the Currency window.
- A new button on the Currency window.
### Fixed
- Incorrect check for Framer's Kits has been resolved.
### Changed
- ECommons is now used as a NuGet package.


philpax:
It looks like you still have `ECommons` as a Git module in your `.gitmodules`, which is stopping Plogon/Git from fetching it. Removing a submodule can be quite irritating, unfortunately, so try playing around with the manual steps here https://stackoverflow.com/questions/1260748/how-do-i-remove-a-submodule

philpax (approved):

PR #6971, opened 2025-09-09
Removed a plugin that was flawed at its core

Pistachio-dev:


philpax:
Er, we're happy to sort this out, but what's the story here?

Pistachio-dev:
I submitted a plugin a while before, but then noticed a fault that made the entire point of the plugin not work. So I want to remove it from the listing

philpax:
bleatbot, approve

PR #6997, opened 2025-09-14
Added Fashion Report to testing and submission

TheRedheadedWitch:
Added Fashion Report mod for testing and submitting as a new plugin


philpax:
bleatbot, approve

philpax:
Hi, our secret system is not really meant to be used for this kind of thing. We *very* strongly advise against connecting to your database directly from the plugin; that is a significant security hazard for both you and your users (as a malicious actor could potentially poison your database in such a way to break your plugin).

Please consider using a backend server. We cannot accept this plugin in its current form.

PR #7017, opened 2025-09-21
Cosmic weapon

TheRedheadedWitch:
New Plugin


TheRedheadedWitch:
Hasel added this feature to Vanilla+ for me after seeing this plugin, so it is no longer needed.

PR #7024, opened 2025-09-24
new plugin Shrinkage

mikelosx:
Made a new plugin. scales the size of characters based on their hp so honestly not sure if it fits the guidelines haha.


MidoriKami:
Suggestion for a subtitle/punchline "It's just cold okay?!"

reiichi001:
bleatbot, approve

mikelosx:
Permission denied? Do I need to change something?

reiichi001:
That's a result of a prior error. The actual cause of the build failure is because your csproj references something that doesn't exist.
https://github.com/mikelosx/Shrinkage/blob/2cf6b06b3b3269c63d6f9ddb125103c0bad799c2/Shrinkage/Shrinkage.csproj#L11

(Remove that whole itemgroup)

perchbirdd:
Hello! We are currently pending rules review from all members still, but at present PAC has concerns with the ability to scale the character above 1.0 size. While the maximum slider goes to 3.0, the player's size maxes at 1.0 *unless* the user brings the minimum size above 1.0. This allows the player to scale larger - which also allows the camera to move out farther (unless my testing is mistaken).

While I can't guarantee pending responses from others will approve the plugin even with these changes, I can say that we will not accept this if it allows the player to become larger or increase the camera's field of view.

mikelosx:
Would limiting the minimum slider to 1.0 and removing the camera changes be enough? or should I cap the maximum size in general?

mikelosx:
ok made some changes, minimum scale has a max of 1 and maximum scale has a minimum of 1.

perchbirdd:
> ok made some changes, minimum scale has a max of 1 and maximum scale has a minimum of 1.

Sorry for the significantly delayed response.

I think there was some disconnect here. The player should not be allowed to become larger than 1. So, the size range should be 0.1 (or however small you want) to 1.0. Min range: 0.1 - 1.0, Max range: 0.1 - 1.0.

Critical-Impact:
Hi, just following up on this, does what perchbirdd said make sense? No rush, just making sure that the plugin is currently blocked until this is addressed.

mikelosx:
Just something I've been meaning to get around to but completely forgot about... anyway, i changed it so the character can't go above a scale of 1.

nathanctech:
Hi,

During internal testing it was pointed out that a user could modify the config file to go above the 1.0 scale limit and it also doesn't disable itself during PvP which it really should be doing. If those can be fixed, I'd say we're good to go with this.

mikelosx:
that is something i didn't even consider, my bad. the max scale is removed entirely and it should be disabled in pvp now

nathanctech:
Looks great, thank you for your patience! Please confirm your Discord username here and ping dalamud-pac for a role to access feedback and the plugin help forum. Thanks!

mikelosx:
discord: jediweegie

PR #7027, opened 2025-09-25
NeatNoter 2.7.4 (Test)

shadowkras:
## Bugfixes

- Fixed the index window transparency being affected by other plugins by accident.
- Re-enabled the editor context menu when right clicking the note, instead of being stuck on the word counter.
- Fixed the "Remove as note overlay" context menu item not showing up on the current active note that was set as note overlay.

## Features

- Added a setting to change the default transparency on the note index. This should give users more control of the index appearance.

If any bug if found, please report at: https://github.com/shadowkras/NeatNoter/issues/


MidoriKami:
bleatbot, rebuild

PR #7037, opened 2025-09-28
BetterFriendList 1.0.2.2 [Stable]

devckuw:
fix :
- pf retrieve data not getting world and private tab
add :
- option for column size
- search pf info if playerId not present in data before trying to open it


philpax (approved):
Sorry for the delay on this; it took us some time to decide if the multiple refreshes counted as one action for the purposes of the automation guidelines.

That being said, we're not sure about the threadsafety of this; from a quick read, it looks like you're calling the refresh function from another thread. I'd suggest scheduling them to be run on the main thread instead.

PR #7163, opened 2025-11-16
BilingualTooltips 1.4.3.0

Elypha:
Hi!

The same fix as https://github.com/goatcorp/DalamudPluginsD17/pull/7162 for the problem in the library


philpax (approved):
fyi you can update multiple plugins in one PR if you'd like

PR #7628, opened 2025-12-31
ZodiacBuddy 1.14.0.0

foophoof:
API14 / 7.4 support


foophoof:
Sorry for the gigantic reformat diff, was hoping the semantic diff would make it a bit easier to read, but looks like that didn't get added here :/

philpax (approved):

PR #7694, opened 2026-01-05
Remove old/non-existent GitHub user from manifests

mabako:


philpax (approved):

philpax:
Hmm, might be cleaner to do this when you have an update for the plugins; I think I can force this through, but Plogon's behaviour around no-op updates is always a bit weird

PR #7854, opened 2026-02-08
Add Character Profile Switch (testing)

ssalazar217:
## Character Profile Switch – Initial submission (testing/live)

**Single plugin per PR:** This PR adds only Character Profile Switch.

**Track:** `testing/live/CharacterProfileSwitch/` (new plugins go to testing as per [submission process](https://dalamud.dev/plugin-publishing/submission)).

### What the plugin does
- Automatically switches the set of enabled plugins when you log in with each character.
- Group characters (Name@World) and assign a list of plugins per profile.
- Use `/cps` to open settings.
- Uses Dalamud Windowing API for config and UI; icon 256×256 in `images/icon.png`; version fixed in .csproj (no timestamp).

### Checklist
- [x] Plugin metadata set (Name, Author, Description, RepoUrl, IconUrl, etc.) in repo manifest.
- [x] [Technical considerations](https://dalamud.dev/plugin-development/technical-considerations) followed (icon size, windowing API, versioning).
- [x] [Restrictions](https://dalamud.dev/plugin-publishing/restrictions) reviewed: no automation of game actions, no combat, no Mog Station; only profile management per character.
- [x] One branch for this submission: `testing/characterprofileswitch`.
- [x] Directory structure: `testing/live/CharacterProfileSwitch/manifest.toml` and `testing/live/CharacterProfileSwitch/images/icon.png`.

Repository: https://github.com/ssalazar217/CharacterProfileSwitch


ssalazar217:
Hi, well, since I started developing it I've been testing and using it all the time, and when I switch characters it performs the exact action I intended when I created it. To date I haven't had any crashes, failures or alerts.

It does generate two warnings when switching between profiles, but I think they're due to the plugins it activates.

image

I can change active groups without changing characters, but the last ones I left active on the character are automatically reactivated when I log back in with the same character without needing to readjust them (which is precisely what I created it for).

That 244ms delay occurs while the active plugins are changing their state.

nathanctech:
Hello,

PAC has decided not to allow this plugin into the official repositories. We have Plugin Collections which allow users to do the similar functions within Dalamud and is a much more stable way to handle it. Thanks for your submission and patience.

PR #7955, opened 2026-03-05
BetterFriendList to 2.0.0.0 testing

devckuw:
add:
    refresh button in native friend list
    collapse button on social (player search / friend list / party members)
    option to prevent friendlist to refresh on open but still keep other refresh (on log in / leaving duty / changing wolrd)
    option to show notes in native friend list (show note instead of the online statue)
    option to keep same keybind for both native and plugin friend list
fix:
    fix refresh when used on specific player

### DIFF
--- a/testing/live/BetterFriendList/manifest.toml
+++ b/testing/live/BetterFriendList/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/devckuw/BetterFriendList.git"
-commit = "f13c7e25acc03ddf9fec59bb62f2fa2223476834"
+commit = "10b421b8b1574be71f5a8f7a7aaaa5bfb10f59a9"
 owners = ["devckuw"]
 project_path = "BetterFriendList"
 changelog = """
--- a/testing/live/BetterFriendList/manifest.toml
+++ b/testing/live/BetterFriendList/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/devckuw/BetterFriendList.git"
-commit = "ca44966f6f48a478e1ab1e70a8540cfd57f390d8"
+commit = "6768dfbff0a8fea08ccb9f2f7973aa9edab8d1eb"
 owners = ["devckuw"]
 project_path = "BetterFriendList"
 changelog = """
--- a/testing/live/BetterFriendList/manifest.toml
+++ b/testing/live/BetterFriendList/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/devckuw/BetterFriendList.git"
-commit = "10b421b8b1574be71f5a8f7a7aaaa5bfb10f59a9"
+commit = "ca44966f6f48a478e1ab1e70a8540cfd57f390d8"
 owners = ["devckuw"]
 project_path = "BetterFriendList"
 changelog = """
@@ -11,6 +11,10 @@ add:
     option to show notes in native friend list
     option to use color in native friend list
     option to keep same keybind for both native and plugin friend list
+    restriction when requesting data (gpose, disconnected, offline, cutscene, afk, camera mode)
 fix:
-    fix refresh when used on specific player 
+    fix refresh when used on specific player
+    fix lodestone page with name with special chars
+    fix icons for crossworld party leader and memeber
+    fix condions for pf button
 """
--- a/testing/live/BetterFriendList/manifest.toml
+++ b/testing/live/BetterFriendList/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/devckuw/BetterFriendList.git"
-commit = "cc620f3c5ff0209109056925f62cb34cad9ba88e"
+commit = "26c9d50930c1e0531d6269d26c6cb9c250ae7453"
 owners = ["devckuw"]
 project_path = "BetterFriendList"
 changelog = """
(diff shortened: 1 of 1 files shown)


MidoriKami:
Diff link for other reviewers, diff from testing to current release 

https://github.com/devckuw/BetterFriendList/compare/e1e0df1a4b5bc3b2d4e65a760bc10a3c2aad9299..3cadc2da29d4e084327f379714f04767bf43343c

MidoriKami:
Happy to see more KTK usage, a few notes.

```cs

        minimizeButton?.DetachNode();
        minimizeButton?.Dispose();

        refreshButton?.DetachNode();
        refreshButton?.Dispose();
```

It is recommended to simply .Dispose(), and then set the variable to null, instead of Detach -> Dispose. Dispose does a more through job of making sure all children are handled correctly.

```cs
    public unsafe void DisableFlagNode(AtkResNode* node, NodeFlags flag)
    {
        node->NodeFlags &= ~flag;
    }

    public unsafe void EnableFlagNode(AtkResNode* node, NodeFlags flag)
    {   
        if (node == null) return;
        node->NodeFlags |= flag;
    }
```

Use KamiToolKit.Extensions.AtkResNodeExtensions AddNodeFlags and RemoveNodeFlags instead

```cs
var windowsResNodeList = windowsResNode->GetComponent()->UldManager.NodeList;
var crossButtonNodeList = crossButton->GetComponent()->UldManager.NodeList;
```

The result of GetComponent from a AtkResNode *must* be checked for null, as it returns null on type mismatch.

Consider looking at `AddonControl` member of some of the addons with sub-addons to control their behavior instead.

devckuw:
Thank you for the feedback.
I'll apply these advice in the next testing version I'm working on right now except if these changes are needed for my PR to be accepted.

devckuw:
I added the required changes in that commit.
I could change the commit tag in the PR but since it contains a new feature on top of the changes I'm not sure its a good idea as its already a large size.
https://github.com/devckuw/BetterFriendList/commit/760f0db04258c43307f68de107e13f01557fa6de

MidoriKami:
Do **not** use `NodeList` by index
```cs
var componentList = friendList->UldManager.NodeList[8]->GetAsAtkComponentList();
```

This *will* crash the game when any modifications are made by anything else.

devckuw:
How would you recommend to get a specific one then?

MidoriKami:
Always use the GetNodeById or GetComponentByNodeId or GetComponentNodeByNodeId and other similar functions. You must null check before each dereference `->` else you risk crashing the game when anything changes.

Some example code from VanillaPlus

```cs
    private void AttachNodes(AddonNeedGreed* addon) {
        var listComponentNode = (AtkComponentNode*) addon->GetNodeById(6);
        if (listComponentNode is null) return;
        
        var listComponent = (AtkComponentList*) listComponentNode->Component;
        if (listComponent is null) return;

        foreach(uint nodeId in Enumerable.Range(21001, 31).Prepend(2)) {
            var listItemRenderer = listComponent->UldManager.SearchNodeById<AtkComponentNode>(nodeId);
            if (listItemRenderer is null) continue;
            
            var listItemRendererComponent =  (AtkComponentListItemRenderer*) listItemRenderer->Component;
            if (listItemRendererComponent is null) continue;

            var targetPart = listItemRendererComponent->UldManager.SearchNodeById(12);
            if (targetPart is null) continue;
```

It's even safer to use the strongly named functions like GetTextNodeById, because that will internally type check that the returned node is the correct type, else it returns null.

nathanctech:
bleatbot, rebuild

nathanctech:
Hello,

Using the InfoProxy is extremely risky here as the client may request data when it's not in the correct state to do so. I see you have basic checks for instances and being logged in, but there are other scenarios where the friends list can not be updated by the game itself, such as during cutscenes (where the UI for the plugin can remain open and interactable) or when the friends list is closed. The plugin also refreshes itself on world changes which is not permitted per our automation rules and is not something the client is capable of doing.

I'd advise making the refresh entirely dependent on opening the native UI and letting the game do the requests as it's very risky sending network requests when the server may not expect them.

devckuw:
Hello,

I can add checks for cutscene and prevent the data request during it. I'm happy to add any more check you can think of to ensure the request data only happen when the friend list can be open.
edit: checking for cutscene, camera, afk, disconnected, not found in InfoProxyDetail and gposing in clientstate

For the auto request they are only here to replace the ones i block. I remove the possibility for the game to request friend list data if the option is set so it doesnt request data when i open the native UI but since it also means it totally block all request for the friend list i request manually data on 3 cases :
- when the player leave an instance
- when the player travel to another world
- when the player log in

these 3 behavior are the only 3 request call done by the game by default whether the native UI is open or not

For the request when the native UI is not open, it alrdy have been discussed and was the main topic when i first asked to add the plugin to main repo and wouldnt have been added if it was a problem.

devckuw:
Most new lines in v2.1.3.5 are just me changing how the main window is drawn plus some tools i use for the development. I pushed them with the fix asked above but i removed their calls.

nathanctech:
Thank you. User safety is one of the biggest things I look for and I want to make absolutely certain that clients with this plugin behave the same way as clients that don't to avoid any issues down the line. The code changes will be reviewed and thank you for your patience.

nathanctech:
After discussion, the plugin will be merged to the testing branch. Please monitor the plugin's behavior to make sure it behaves as you expect it to and does not conduct server requests when it's not supposed to.

nathanctech (approved):

devckuw:
Yes, I try to monitor my plugins' behavior as much as i can so if i see anything suspicious I'll resolve it as fast as i can!

PR #7958, opened 2026-03-06
[New] WukLamark 1.0.2.2 (formerly WukWaymark)

Bronya-Rand:
**Initial release of WukLamark to 'testing/live'**

## What's New
- Create and manage custom waymarks that display visibly on both the Main Map and Minimap.
- Highly customizable markers: Use simple shapes (with custom colors) or pick from hundreds of in-game icons.
- Groups: Organize your waymarks into nested, collapsible categories.
- Waymark/Group Scopes: Designate waymarks and/or groups as "Personal" (locked to your specific character) or "Shared" (visible to all characters that log in on your PC).
- Read-Only Waymarks/Groups (for Shared Waymarks/Groups): Protect shared waymarks and groups as "Read-Only" so you (or someone else) don't accidentally edit or delete them!
- Import/Export: Easy clipboard sharing to distribute groups of waymarks to your friends.


reiichi001:
bleatbot, approve

MidoriKami:
Code looks mostly good, [40a43a0](https://github.com/Bronya-Rand/WukWaymark/commit/40a43a09baa40053323096cb01d2d1e8f3bef15b) couple minor remarks, 

The minimap overlay may be missing ImGuiHelpers.GlobalScale modifiers, this might be a problem when non 100% scale is used.

`WaymarkMinimapWindow.cs` *needs* to have null checks on the node getters, while it's unlikely that the node ID's will change, it can happen, and would cause instant crashing if it did happen.

Bronya-Rand:
> Code looks mostly good, [40a43a0](https://github.com/Bronya-Rand/WukWaymark/commit/40a43a09baa40053323096cb01d2d1e8f3bef15b) couple minor remarks,
> 
> The minimap overlay may be missing ImGuiHelpers.GlobalScale modifiers, this might be a problem when non 100% scale is used.
> 
> `WaymarkMinimapWindow.cs` _needs_ to have null checks on the node getters, while it's unlikely that the node ID's will change, it can happen, and would cause instant crashing if it did happen.

Updated hash to 1.0.1 which fixes (I hope all) these things + more.

Bronya-Rand:
Recent PR merged making another bump to 1.0.1.1

Bronya-Rand:
Fair warning to those testing WukWaymark 1.0.2.0 who tried 1.0.0.0-1.0.1.1.

Due to the newly added permission system, obfuscation, etc. you MUST delete the WukWaymark.json file and any files in the `WukWaymark` folder. Since this has not yet merged, I didn't really consider migration only until it is on `testing/live`.

New stuff in WukWaymark 1.0.2.0:
- Waymark Groups
- Game Icon support
- Caching and optimizations
- New alias command and `/wwmark here <Group>`
- Waymarks per character
   > ContentId is hashed.

- Fade effect for markers at edge of map.
- Hide markers after X yalms
- Table/Group views
- Import/Export waymarks

Probably a few more but those are the main ones.

foophoof:
Reviewed https://github.com/Bronya-Rand/WukWaymark/commit/f183dfbf1b99f1e246c24d80dabe1f037c1994b8

All looks fine safety-wise, but a few things I noticed from reading through the code and testing it:

(These are generally fairly minor, and nothing I personally think should block this PR, think of them more as…early testing feedback 😅)

- I reversed a bit more of `NaviMap` recently, so you can get some of the NaviMap state without having to traverse the Atk component tree now: https://github.com/aers/FFXIVClientStructs/pull/1743. What you have now works fine, and the CS change is in `stg` only right now, but mentioning it in case you want to get rid of some code in the future.
- https://github.com/Bronya-Rand/WukWaymark/blob/f183dfbf1b99f1e246c24d80dabe1f037c1994b8/WukWaymark/Windows/WaymarkWindow.cs#L60 - `EndTooltip` should only be called if `BeginTooltip` returns true (see [imgui header for docs](https://github.com/ocornut/imgui/blob/5dd56d4bb5e0c0f0dbb317857157940188898853/imgui.h#L824)). You should also check the return value of `BeginTooltip` before writing text. I'd recommend using ImRaii instead:
  ```csharp
  if (…)
  {
      using var tooltip = ImRaii.Tooltip();
      if (tooltip)
      {
          ImGui.TextUnformatted(…);
      }
  }
  ```
  There's a few other places you're calling `ImGui.EndTooltip` without checking the return value from `BeginTooltip`, not just the one I linked to. Or, alternatively, switch to `ImGui.SetTooltip` if you only need to set some plain text.
  (I'd also generally recommend using ImRaii anywhere you have an `ImGui.Begin…` or `ImGui.Push…`, makes it harder to accidentally not call `End…` / `Pop…` when you need to).
- The NaviMap markers appear to ignore whether the minimap is locked north or not (it's only in the right position if I lock the minimap north to be up, if I unlock it the marker stays in the same screen position, which is now wrong unless I'm facing north).
- The items to the right of the searchbar in the main window (current zone checkbox and undo button) seem to assume that the Dalamud font scale is 100% (or less), if you increase the font scale they start to drift outside the window.
- The markers on the map end up on top of other plugin windows that are over the map, you may want to consider `GetBackgroundDrawList` instead of `GetForegroundDrawList`.
- You have a few cases of passing an `IEnumerable<Waymark>` to a function, and then you iterate over it multiple times. This appears to work fine in all the cases I looked at, since you're actually passing a class, but be aware that not all `IEnumerable<T>`s can be iterated over multiple times, so if you ever end up changing what concrete type you are passing in the future, you may see some bugs appear without the type system helping you.

Bronya-Rand:
> Reviewed [Bronya-Rand/WukWaymark@f183dfb](https://github.com/Bronya-Rand/WukWaymark/commit/f183dfbf1b99f1e246c24d80dabe1f037c1994b8)
> 
> All looks fine safety-wise, but a few things I noticed from reading through the code and testing it:
> 
> (These are generally fairly minor, and nothing I personally think should block this PR, think of them more as…early testing feedback 😅)
> 
>     * I reversed a bit more of `NaviMap` recently, so you can get some of the NaviMap state without having to traverse the Atk component tree now: [Add some NaviMap stuff aers/FFXIVClientStructs#1743](https://github.com/aers/FFXIVClientStructs/pull/1743). What you have now works fine, and the CS change is in `stg` only right now, but mentioning it in case you want to get rid of some code in the future.
> 
>     * https://github.com/Bronya-Rand/WukWaymark/blob/f183dfbf1b99f1e246c24d80dabe1f037c1994b8/WukWaymark/Windows/WaymarkWindow.cs#L60 - `EndTooltip` should only be called if `BeginTooltip` returns true (see [imgui header for docs](https://github.com/ocornut/imgui/blob/5dd56d4bb5e0c0f0dbb317857157940188898853/imgui.h#L824)). You should also check the return value of `BeginTooltip` before writing text. I'd recommend using ImRaii instead:
>       ```cs
>       if (…)
>       {
>           using var tooltip = ImRaii.Tooltip();
>           if (tooltip)
>           {
>               ImGui.TextUnformatted(…);
>           }
>       }
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       There's a few other places you're calling `ImGui.EndTooltip` without checking the return value from `BeginTooltip`, not just the one I linked to. Or, alternatively, switch to `ImGui.SetTooltip` if you only need to set some plain text.
>       (I'd also generally recommend using ImRaii anywhere you have an `ImGui.Begin…` or `ImGui.Push…`, makes it harder to accidentally not call `End…` / `Pop…` when you need to).
> 
>     * The NaviMap markers appear to ignore whether the minimap is locked north or not (it's only in the right position if I lock the minimap north to be up, if I unlock it the marker stays in the same screen position, which is now wrong unless I'm facing north).
> 
>     * The items to the right of the searchbar in the main window (current zone checkbox and undo button) seem to assume that the Dalamud font scale is 100% (or less), if you increase the font scale they start to drift outside the window.
> 
>     * The markers on the map end up on top of other plugin windows that are over the map, you may want to consider `GetBackgroundDrawList` instead of `GetForegroundDrawList`.
> 
>     * You have a few cases of passing an `IEnumerable<Waymark>` to a function, and then you iterate over it multiple times. This appears to work fine in all the cases I looked at, since you're actually passing a class, but be aware that not all `IEnumerable<T>`s can be iterated over multiple times, so if you ever end up changing what concrete type you are passing in the future, you may see some bugs appear without the type system helping you.

All points have been noted and Points 2-5 have been addressed. Same version as this is minor stuff that doesn't affect the overall update code logic.

Bronya-Rand:
Slight notice for 1.0.2.1 for PAC and Code Reviewers: Delete all existing waymarks. As I added Ward identification support, the new `WardId` field will be missing for old waymarks and will be designating everything as "New Gridania - Ward 1" for some reason. Only reason I haven't set it to `sbyte? WardId` is because this has yet been merged.

Bronya-Rand:
Okay so update log here:
1.0.2.2 is mostly a UI adjustment but I also removed the obuscation part as Midori told me as long as the data is local, said stuff is fine to keep in plaintext. Only thing that has been kept is a SHA hash.

Existing waymark data will not work and will have to be deleted due to this change.

nathanctech:
Marking as needing review to go over the changes made.

nathanctech (approved):

PR #7976, opened 2026-03-12
Delete PartySortPlus for now until it's fully ready again

tenyu27:
Notice of adoption: https://discord.com/channels/581875019861328007/653504487352303619/1473109458224812062


Infiziert90:
The plugin should just be banned (DalamudAssets repo) instead of deleted. 
Deleting it may have some weird side effects

reiichi001:
@Infiziert90 is correct. If the plugin currently is API 14, and failing to load, you can have it blocked from loading by adding to the `bannedplugin.json` list in https://github.com/goatcorp/DalamudAssets . 

If it's out of date already, there's no need to do so.

When you are ready to release an update, you'll want to adjust the manifest.toml file.
- If you have full ownership, replace Rietty on the owners field.
- If you're sharing ownership, add yourself there.
- If Rietty stays the owner, but you're taking over maintenance, add a `maintainers` array, like the owners one. https://discord.com/channels/581875019861328007/1061004400023322792/1457161649038430455

PR #8007, opened 2026-03-18
Add ChatAnywhere to testing/live

twelvehouse:
## ChatAnywhere

Adds ChatAnywhere to the testing/live track.

**Plugin summary:** Runs a lightweight web server inside FFXIV that streams in-game chat to any browser in real time, and allows sending messages back without switching to the game window.

**Key features:**
- Real-time chat streaming via Server-Sent Events (major chat channels and selected system/emote logs; battle log channels are intentionally excluded to avoid flooding the history)
- Custom filters with drag-and-drop reordering
- Tell mode with pinning and channel switcher (recent DMs sidebar is UI-only placeholder, not yet implemented)
- FFXIV special character picker
- URL link previews (YouTube, Lodestone, etc.)
- Appearance settings (Google Fonts, font size, coloured backgrounds)

**Notes:**
- Inspired by ChatTwo's web frontend concept; written from scratch as a standalone plugin
- Frontend (React/Vite) is pre-built and committed as `websiteBuild.zip` for Plogon compatibility (`IsPlogonBuild=True` skips the npm step and extracts the zip instead)


reiichi001:
bleatbot, approve

yfthbr:
A couple concerns as of [bf26eaa](https://github.com/twelvehouse/ChatAnywhere/commit/bf26eaabd40857de65084f130519d16653737241)

- Lack of any kind of auth on the API endpoints. You're opening a port that is open to the world barring any firewall configuration the user may or may not have. It is not safe to assume only the user will be able to make requests to the web server in the plugin. Either add some mechanism to ensure only the authorized user can use the API or make the web server only listen to localhost.
- CORS explicitly allows any website to call the settings and send message endpoints. Ideally it would be scoped to only the origin that serves the front, but not a deal-breaker if the first point is handled with strong auth.
- HandleSendMessage allocates a buffer based on client headers with no limit. Especially linked with above, it could be used to make the game and system run out of memory by someone malicious. Since the messages are already limited to 500 bytes, it would be good to limit the buffer to a reasonable number based on that, accounting for overhead from the overall json model.
- HandlePutSettings has the same issue.
- ChatReceiver is marked as IDisposable but WebServer doesn't dispose of it.

Frontend code seems ok.

twelvehouse:
> A couple concerns as of [bf26eaa](https://github.com/twelvehouse/ChatAnywhere/commit/bf26eaabd40857de65084f130519d16653737241)
> 
> * Lack of any kind of auth on the API endpoints. You're opening a port that is open to the world barring any firewall configuration the user may or may not have. It is not safe to assume only the user will be able to make requests to the web server in the plugin. Either add some mechanism to ensure only the authorized user can use the API or make the web server only listen to localhost.
> * CORS explicitly allows any website to call the settings and send message endpoints. Ideally it would be scoped to only the origin that serves the front, but not a deal-breaker if the first point is handled with strong auth.
> * HandleSendMessage allocates a buffer based on client headers with no limit. Especially linked with above, it could be used to make the game and system run out of memory by someone malicious. Since the messages are already limited to 500 bytes, it would be good to limit the buffer to a reasonable number based on that, accounting for overhead from the overall json model.
> * HandlePutSettings has the same issue.
> * ChatReceiver is marked as IDisposable but WebServer doesn't dispose of it.
> 
> Frontend code seems ok.

Thank you for the thorough review. All points have been addressed:

**Auth:** All API endpoints now require a session cookie (`chat-anywhere-token`). (Static file serving — `/`, `/files/*`, `/icon/{id}` — remains public as these contain no user data.) The user sets a 4–8 digit numeric passcode in the plugin settings UI; it is stored as a PBKDF2-SHA256 hash (random 16-byte salt, 100k iterations). On `/auth POST`, the submitted passcode is verified with `CryptographicOperations.FixedTimeEquals` to prevent timing attacks, and a random UUID token is issued as an `HttpOnly; SameSite=Strict` cookie valid for 30 days. The frontend shows a passcode modal on load and re-prompts if any request returns 401 mid-session.

**CORS:** The handler now reflects the request's `Origin` header instead of sending a wildcard. This still allows any origin to receive a credentialed response, but `SameSite=Strict` on the session cookie means browsers will not attach the cookie to cross-site requests, so cross-origin CSRF is not exploitable in practice. Scoping to a fixed origin whitelist is not straightforward here because the server port is user-configurable, so we consider this an acceptable trade-off given the cookie protection.

**Buffer limits:** `HandleSendMessage` now reads at most 4 096 bytes via a shared `HttpHelper.ReadBodyAsync` utility that respects `Content-Length` and never reads past the declared size. `HandlePutSettings` is capped at 65 536 bytes. Both return 400/413 if the limit is exceeded.

**ChatReceiver disposal:** Fixed. `_receiver.Dispose()` is now called inside `WebServer.DisposeAsync()`.

yfthbr:
As of [1227bd1](https://github.com/twelvehouse/ChatAnywhere/commit/1227bd1db0c490a24b88d3356a801bbd778c77cc)
- OgpHttpClient is not disposed?
- Core/SseManager.cs:51 can end up blocking the game thread for up to 500ms upon receiving a message, which is no good (unless I missed something) as it seems to bubble up directly from ChatGui.ChatMessageUnhandled handler in some cases. If it needs to be waited, it should happen off of the main game thread.

yfthbr:
code seems good as of [4dbbd94](https://github.com/twelvehouse/ChatAnywhere/commit/4dbbd94fd3e93dc940b4b1a959c1b82f4a73b08b)

twelvehouse:
> As of [1227bd1](https://github.com/twelvehouse/ChatAnywhere/commit/1227bd1db0c490a24b88d3356a801bbd778c77cc)
> 
> * OgpHttpClient is not disposed?
> * Core/SseManager.cs:51 can end up blocking the game thread for up to 500ms upon receiving a message, which is no good (unless I missed something) as it seems to bubble up directly from ChatGui.ChatMessageUnhandled handler in some cases. If it needs to be waited, it should happen off of the main game thread.

Thanks for the review. Both points have been addressed:

**OgpHttpClient disposal:** Fixed. `OgpHandler` now implements `IDisposable`; `_httpClient` is an instance field disposed via `WebServer.DisposeAsync()`.

**SseManager game-thread blocking:** Fixed. `Broadcast` now enqueues a `byte[]` into a `Channel<byte[]>` via `TryWrite` and returns immediately, so the caller (including the game thread) is never blocked. A single consumer task (`ConsumeAsync`) drains the queue sequentially on the thread pool, which also guarantees event delivery order. `SseManager` is now `IAsyncDisposable` — shutdown calls `Writer.Complete()` and awaits the consumer task to ensure clean termination.

nathanctech:
@twelvehouse We have recently adopted a new [AI Usage Policy](https://github.com/goatcorp/governance/blob/main/ai-policy-official-repo.md) for all plugins submitted to the official repository. Please give it a read and update your plugin description and pull request message if necessary.

twelvehouse:
> @twelvehouse We have recently adopted a new [AI Usage Policy](https://github.com/goatcorp/governance/blob/main/ai-policy-official-repo.md) for all plugins submitted to the official repository. Please give it a read and update your plugin description and pull request message if necessary.

Thanks, I've read through the policy. Here's my disclosure.
- Code: Copilot level — architecture and decisions are mine; all output reviewed
- Comments / commit messages / PR text: Assist level (non-native English speaker)
- Icon: AI-generated (Gemini). I can remove it if that's a problem.

nathanctech:
We would prefer a non-AI icon, yes. These are displayed to users in the Plugin Installer. As you can tell from other plugins that exist, they don't have to be fancy or by an artist. :)

twelvehouse:
Sure, I removed the ai generated icon.

yfthbr:
code seems good as of [097474e](https://github.com/twelvehouse/ChatAnywhere/commit/097474ef9b2ca8334482d9665b185eaea83db844)

nathanctech:

nathanctech:
@twelvehouse FYI, I recommend updating your TOML changelog to include the plugin's description and such for the first release as that's what will get posted. Just waiting for internal testing and we'll be good to go.

Infiziert90:
Gave your plugin a quick look again and found some issues that should be fixed before releasing to the public.

- Don't pass byte arrays to CS without a null terminator at the end
https://github.com/twelvehouse/ChatAnywhere/blob/main/Core/ChatSender.cs#L41
- It would be nice if code taken from ChatTwo (looking at ChatSender class) would have been marked as such :)
- "*" is invalid in newer versions of Watson, bind to 0.0.0.0 instead
- Your plugin should make heavy use of the newly added `Async` plugin class, as its very much async focused
- Your .json file contains a lot of invalid values that aren't supposed to be used on main repo plugins, it shouldn't contain much more than what SamplePlugin exposes
https://github.com/goatcorp/SamplePlugin/blob/master/SamplePlugin/SamplePlugin.json
- `ImGui.SameLine(80);` hardcoded values like this should be scaled with `* ImGuiHelpers.GlobalScale` or users with different HUD Scale won't have a good time
- `GetCurrentGameChannelPrefix()` "/n" is not a valid prefix for other languages, revert it to /b or /beginner


Did not check any of the Frontend code

Infiziert90:
IconUrl and the other one are also handled by DP17 and shouldn't be used in your json file

With hardcoded values I didn't mean just `SameLine()` but all hardcoded values, so also the `SetNextItemWidth` one for example
if you have height / width values hardcoded, scale them by Globalscale, only exception is Window.Size, this one gets scaled by dalamud

twelvehouse:
@nathanctech @Infiziert90 
Thank you for the follow-up reviews.

- Null terminator: `SendMessageUnsafe` now uses `[.. message, 0]` with `Utf8String.FromSequence`
- ChatTwo attribution added at the top of `ChatSender`
- Watson binding changed to `"0.0.0.0"`
- Migrated to `IAsyncDalamudPlugin` — startup and event hooks moved to `LoadAsync`, disposal now properly awaits `Server.DisposeAsync()` instead of blocking
- Cleaned up `ChatAnywhere.json` — removed `DalamudApiLevel`, `LoadRequiredState`, `LoadSync`, `LoadPriority`, `CanUnloadAsync`, `IconUrl`, and `ImageUrls`
- `ImGuiHelpers.GlobalScale` applied to all hardcoded pixel values in `SettingsWindow` (both `SameLine` offsets and `SetNextItemWidth` calls)
- NoviceNetwork prefix fixed (`/b`)

nathanctech:
@twelvehouse Thanks for your patience, your plugin has been merged! Please keep an eye on the testing channels in case our testers find something. Please note your discord name here.

Please note to keep us informed if your external CI updates the website build as this isn't easily trackable by PAC and all frontend code changes should also be PRed like usual.

nathanctech (approved):

twelvehouse:
@nathanctech 
Thank you for merging! My Discord name is `laincandy`.

Regarding the website build: my CI automatically commits the built websiteBuild.zip directly into the repository, and I always use that commit hash in manifest.toml. So the build state is fully tied to a specific commit and should be trackable as-is.

PR #8013, opened 2026-03-19
Add PullToOBS plugin

Miu-B:
PullToOBS talks to OBS over WebSocket v5, starts recording the moment you enter combat, and saves a replay buffer clip of the prepull.

**AI Usage:** Copilot — I have programming experience, though not in C# or the Dalamud API specifically. I defined the requirements, made architectural and design decisions, reviewed all code for patterns and logic, and tested the plugin in-game. The C# and Dalamud-specific code was written by an AI assistant.

**Features**

- Starts OBS recording when combat begins, stops after a 5-second grace period
- Re-entering combat during grace period cancels the pending stop for continuous recording
- Saves replay buffer 5 seconds into the encounter to capture the prepull
- Visual on-screen indicator (red pulsing = recording, orange = replay buffer active, green = connected, grey = disconnected)
- Configure via `/pulltoobs` or `/pto`; toggle OBS connection with `/pto obs`

**Testing**

- Tested in-game with OBS WebSocket v5
- Verified recording starts/stops with combat transitions
- Verified replay buffer save on encounter start
- Verified grace period cancellation on combat re-entry

**Technical Details**

- Uses Dalamud Windowing API for all windows
- OBS calls run off the game thread (ThreadPool/kernel timers)
- No server interactions beyond OBS WebSocket


Infiziert90:
Just gave your plugin a quick look and noticed two things,

Please do not write every single log line of yours to the Information channel, it will clog up the log long term, especially if it does so often, most of these lines are more for Debug or even Verbose.

Also please do not spawn a window inside a window, just use dalamuds window class, span the whole Viewport, and draw or
you are smart and draw directly in your `UiBuilder.Draw()` function, as you don't even need a window for that at all

Miu-B:
Thanks for the review! Both points are fair.

For the logging, I added verbose diagnostics to help debug a production issue (recording not stopping reliably) — that's now fixed, so I'll strip it back to just the essential lifecycle events.

For the indicator, I'll refactor it to a proper Dalamud `Window` subclass spanning the full viewport instead of calling `ImGui.Begin()` manually.

Will push an update shortly.

Miu-B:
Addressed both points:
- **Logging**: Demoted all operational logs (combat state changes, start/stop recording, replay buffer, timers, etc.) from `Information` to `Debug` across `EncounterManager` and `OBSController`. Only 4 `Information`-level lines remain: connecting, connected, disconnecting, and plugin loaded.
- **Window-inside-a-window**: Refactored `OBSStatusIndicator` to extend Dalamud's `Window` class, registered with `WindowSystem`. Uses `PreDraw`/`Draw`/`PostDraw` overrides - removed the manual `ImGui.Begin()`/`ImGui.End()` entirely.

Infiziert90:
You can use `DrawCondition()` to prevent drawing, this will also prevent `PreDraw` and `PostDraw` so you would also need to check draw conditions in `Update()` instead.

Miu-B:
Great suggestion, thanks! I'll check out how to use `DrawConditions()` instead.

nathanctech:
@Miu-B We have recently adopted a new [AI Usage Policy](https://github.com/goatcorp/governance/blob/main/ai-policy-official-repo.md) for all plugins submitted to the official repository. Please give it a read and update your plugin description and pull request message if necessary.

Miu-B:
Updated the PR description to include an AI usage disclosure as requested.

nathanctech:
Thanks, the code looks fine since it doesn't heavily rely on Dalamud to function at least. Just needs to be tested and this should be good to go. Thanks for your patience.

nathanctech:
@Miu-B Sorry for the delay on testing, finally was able to. It appears to not use the replay buffer as intended and only starts recording when you enter combat, with it only posting a single file from the start of combat until the end (where it stops about 5 seconds after). Based on your description, it should also be recording the pre-pull. I attempted to use the countdown timer as well to trigger it and it didn't happen, so I'm not sure how that works.

Miu-B:
@nathanctech thanks for your time and feedback!

The only requirement is having the **OBS Reply Buffer** configured. Here's my setup for reference:

image

**How it works**

Once connectet, the status indicator will turn **Orange**, confirming the Replay Buffer is active ready to record. 

image


When an encounter begins:

1. Recording starts immediately
2. After 5 seconds, it saves the last X seconds from the Replay Buffer (30 seconds in my configuration)
3. This creates an overlap between the two clips

I use this daily on Linux and have had friends successfully test it on Windows as well.

Clearly my instructions aren't clear enough. I'll work on improving them. Let me know if anything else needs clarification!

nathanctech:
Ah, I see now. That makes more sense. It seems to work just fine otherwise, feel free to add any clarification you think it needs and submit an update. We're good to go here. If you're on the Dalamud discord, please note your discord name here and ping dalamud-pac to get access to some dev channels.

nathanctech (approved):

Miu-B:
Hey @nathanctech - Thanks for the quick check and the merge! My discord account "miu_bittermoon" 
Not clear who should I ping though ^^

PR #8061, opened 2026-04-02
[Add] DutySpeed

EuphoricCrashOut:
DutySpeed is a lightweight, automated "speedrun" timer for Final Fantasy XIV. It tracks the time it took to complete dungeons, trials, and raids without needing to manually start or stop a stopwatch. This would be my first submission attempt! \o/


nathanctech:
bleatbot, approve

nathanctech:
Ah, Plogon expects an actual commit hash, not the branch name. Sorry about that.

MidoriKami:
```cs
        if (DutyState.IsDutyStarted && !IsRunning) StartDuty();
        if (!DutyState.IsDutyStarted && IsRunning) EndDuty();
```

IDutyState has a DutyStated and DutyCompleted events you can directly subscribe to, you should use the events instead of polling it every frame.

Infiziert90:
Gave your plugin a look and found these few things that could be changed 

- You uploaded the obj folder, should just delete it

```cs
var territory = DataManager.GetExcelSheet<Lumina.Excel.Sheets.TerritoryType>()?.GetRow(ClientState.TerritoryType);
var name = territory?.PlaceName.Value.Name.ExtractText() ?? "Unknown Duty";
```
- The null checking here is not needed, basic sheets should never be null or dalamud would have massive issues, so writing `GetExcelSheet<TerritoryType>()!` to tell the compiler that this will never be null is fine
- `ExtractText` can now just be replaced with `ToString()`
- `SizeConstraints` and `AlwaysAutoResize` don't work well together, with the Max setting, so should just remove the `SizeContraints` or make the Max be `float.MaxValue`
- Even for very small plugins, it is recommended to use `ImRaii` instead of `ImGui.Begin....`
- `Any()` checks should be avoided because of the performance impact they have, with a list of 5 its not that big, but much simpler is just doing `.Length != 0` 
- Your json manifest contains many bad values, try to copy https://github.com/goatcorp/SamplePlugin/blob/master/SamplePlugin/SamplePlugin.json

MidoriKami:
Also worth noting, the IDutyState service is being updated to include ContentFinderCondition (Duty Name), with some other useful information directly, so you won't even need to look it up.

EuphoricCrashOut:
1.3.2 ---  Updated the plugin to API 14 better standards. Swapped to ImRaii, optimized checks, removed silly package references, and fixed window resizing constraints. Ready for another look! (I think)

nathanctech:
You'll need to update this PR with the new commit hash so it gets picked up.

EuphoricCrashOut:
Heyas! Wanted to provided some info/details ♥ I've been running this personally for about two months or so?? My friends asked about it and then they wanted to try it...  they enjoyed it and thought it was neat, and now here we are because they told me I should submit it. xD (I'm generally super shy) Thanks for your patience with me, and help!

foophoof:
Reviewed EuphoricCrashOut/DutySpeed@79e56e614d6cec75fe3b0041f64f26cce8c9bebd (`79e56e614d6cec75fe3b0041f64f26cce8c9bebd`).

Code looks safe afaict.

As Infiziert90 [mentioned earlier](https://github.com/goatcorp/DalamudPluginsD17/pull/8061#issuecomment-4174190754), your DutySpeed.json file has a lot more things set than it should or needs (e.g. `DalamudApiLevel` and `ApplicableTarget`). I would remove everything except `Author`, `Name`, `Description`, `Punchline`, `CategoryTags`, `Tags` and `RepoUrl`. The rest are all the default values or set automatically by the build tooling and could cause some issues as the tooling changes. If you prefer, you can instead set these directly in your `.csproj` file instead of in a separate JSON file as shown in this Dalamud v14 changelog entry: https://dalamud.dev/versions/v14#csproj-manifest-properties, but that's up to you.

EuphoricCrashOut:
Thanks for all the feedback and kindness! I really appreciate it, and this whole experience ♥♥

nathanctech:
@EuphoricCrashOut Just needs a quick adjustment to your manifest and we should be good to merge!

EuphoricCrashOut:
I'm horrible at Github, so I took a guess it was just the hash that needed to be updated? 😅  Thanks!

nathanctech:
@EuphoricCrashOut thanks for your submission! Please note your Discord username here and ping dalamud-pac to have your role added. :)

nathanctech (approved):

PR #8081, opened 2026-04-10
Glance: new plugin

rphubco:
New plugin submission for testing/live

Glance is the in-game companion for rphub.co, a community-run FFXIV roleplay site (profiles, communities, events). When you target or hover another player it shows their RPHub profile. Basically TRP3 for XIV if you're coming from WoW.

Features:
- Profile tooltip on click/hover
- Create and edit your profile from inside the plugin (rphub.co account only needed to sign in)
- Console and web-only players can create profiles on rphub.co without ever installing the plugin, so a visible profile does not imply plugin usage.
- Tints nameplate icons for players who have a Glance profile
- Optional nameplate name/title and chat name replacement
- Local notes and per-player mute
- Combat/duty hide, "only show roleplayers", plus other display toggles

Source: https://github.com/rphubco/glance
Plugin page: https://rphub.co/glance
Commit: 756ecebab856535e2434fede9de0e5991b5556ae
License: AGPL-3.0
Privacy policy: https://rphub.co/privacy-policy
ToS: https://rphub.co/terms-of-service

Distribution history: we've been shipping through a self-hosted pluginmaster (https://rphub.co/dalamud/pluginmaster.json) since late 2025. Around 4,900 registered accounts, 900 daily unique plugin users, peaks around 400 concurrent. Moving to the official repo so users don't need a custom source anymore.

Content: RPHub is SFW only. Uploads go through automated NSFW/illegal-content screening and we have an active mod team on reports. Anything the plugin renders in-game (profile text, images) is pulled from the site and has already been through that pipeline.

Data handling:

- Auth: per-user API key tied to the RPHub account, stored locally, revocable from account settings.
- Character linking: plugin sends character name, home world, and a SHA-256 hash of the Content ID, computed locally. The raw Content ID never leaves the machine. Ownership is verified with a code placed in the Lodestone profile.
- Profile lookups: if you target or hover someone who isn't already in the Nearby cache, the plugin does a one-off lookup by name + world. Only fires on user action, no passive scanning.
- Nearby Players (opt-in, off by default): first run shows a prompt explaining what gets sent. If enabled, the plugin broadcasts your own name, world, and current zone (or housing plot ID) once per minute. No coordinates, no position, no movement. The server replies with other Glance users nearby so the plugin can tint their nameplates without having to upload every visible player's name. Data is purged 2-3 minutes after disable or logout. Toggleable in settings any time. Existing users who enabled this on earlier versions aren't re-prompted.
- The plugin doesn't collect or transmit data about other players beyond what RPHub users have voluntarily published themselves.
- Plugin user exposure: the Nearby Players list only includes accounts that have explicitly opted into the ingame beacon. The broader profile directory on rphub.co is open to console and web users who have never touched the plugin, so the presence of a profile is not a signal that someone is running Glance.


nathanctech:
bleatbot, approve

Infiziert90:
Gave your plugin a quick look and found some concerning ImGui issues that should be fixed as they will very likely lead to user crashing to desktop

- CachedProfilesTab contains double `EndChild()` which is very much not allowed, to prevent such issues please use ImRaii
- Instead of calling FontScale everywhere (which does some very basic upscaling of the existing atlas) it is better to create your own fonts with Sizes that you like, as an example see: https://github.com/Infiziert90/ChatTwo/blob/main/ChatTwo/FontManager.cs or other plugins that do similar things
- In the nature of C# you really shouldn't have to reflect your own plugins class, just share them via Ctor or other injection methods `var field = typeof(CacheService).GetField("_profiles", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);` 
- Your CSPROJ file should follow https://github.com/goatcorp/SamplePlugin/blob/master/SamplePlugin/SamplePlugin.csproj in the use of Dalamud.Sdk and other fields it sets, anything more isn't necessary 
- Same goes for your json manifest which sets a lot of fields that are handled by the packager, see https://github.com/goatcorp/SamplePlugin/blob/master/SamplePlugin/SamplePlugin.json
- Your own plugin version is readable through `PluginInterface.Manifest` instead of reflecting yourself `typeof(Globals).Assembly.GetName().Version?.ToString(3)`
- `Globals.Data.GetExcelSheet<Lumina.Excel.Sheets.TerritoryType>()?` being null for standard sheets is almost impossible, and if then dalamud won't be running at all, so using `!` is absolutely fine, on the other hand the `GetRow(id)` should be replaced with `TryGetRow()` if the value is not trustworthy
- All calls to `ExtractText()` can also be replaced with `ToString()` since the behaviour in Lumina was changed months ago
- Hardcoded size values won't work the moment users have a different HUD scale value, so `new Vector2(80, 26)` should be replaced with `new Vector2(80, 26) * ImGuiHelpers.GlobalScale`, this is true for all hardcoded value, not things like `ImGui.CalcTextSize`
-  Our ImGui Binding does not accept `Length` as argument anymore, this was only valid for the old ones, instead the last `int` now controls Height, which may not be wanted `ImGui.Combo("##race", ref _raceIdx, RaceData.Races, RaceData.Races.Length)`
- Dalamud already has an icon browser under `xldata icon` which user can freely use, should be used over `DrawIconTabs` as this has no real error handling
- `Globals.PlayerState.CurrentWorld.Value.Name.ExtractText();` and `...HomeWorld` should be used with `ValueNullable` as these are not bound to `LocalPlayer` being null or `PlayerState.IsLoaded`
- Please use `Dalamud.Utility.Util.OpenLink("LINK");` for opening browser links, as it includes error and window handling

rphubco:
Thanks for the review, these are really helpful and would save me a lot of headache if I had researched properly.

Fixed in the push: double EndChild, the _profiles reflection (which turned out to be dead code, GetAllCached was added later and the loop was already using it), version via PluginInterface.Manifest, csproj slimmed down to Dalamud.NET.Sdk, json manifest stripped to match the sample, ValueNullable on PlayerState.CurrentWorld and HomeWorld, TryGetRow on TerritoryType and the Combo Length arg. Also switched to the Dalamud.Utility.Util.OpenLink.

A few I'd like to push back on or do as a followup:
ImRaii everywhere is a big sweeping change and I'd rather not bundle it with this PR. The EndChild bug is fixed directly. I'll do an ImRaii migration as it sown pass after this lands. 
Same for the FontManager and GlobalScale work, they're both real refactors that touch on a lot of files and I'd prefer ship them as dedicated submissions so the diffs stay reviewable. ExtractText to ToString i'll roll into the same cleanup pass

For DrawIconTabs I'd like to keep the picker that I have. xldata icon is a debug and, well, it looks like it. Making users leave the plugin to struggle with the UI only to copy an icon ID back is a noticeable UX hit especially since picking icons is part of the main character setup flow. Happy to tighten the error handling there if you can point at what you'd want to see.

KazWolfe:
If you haven't already, I'd also suggest you join the [Dalamud & XIVLauncher Discord](https://discord.gg/holdshift). We might be able to provide assistance and appropriate tooling for this and other things that come with plugin development.

nebel:
You may be able to eliminate `NameplateNodeService` by accessing the updating nameplate objects via the NamePlateGui handler. For example:

```cs
var namePlateObj = (AddonNamePlate.NamePlateObject*)handler.NamePlateObjectAddress;
```

This should let you access `namePlateObj->NameIcon` to set/reset the tint without all the low-level work `NameplateNodeService` currently does.

nathanctech:
@rphubco We have recently adopted a new [AI Usage Policy](https://github.com/goatcorp/governance/blob/main/ai-policy-official-repo.md) for all plugins submitted to the official repository. Please give it a read and update your plugin description and pull request message if necessary.

nathanctech:

rphubco:
SDK bumped to 15.0.0, with the IChatGui IHandleableChatMessage rewrite, and finally I was able to finish the ImRaii sweep (I really appreciate the conditional PushColor over load), ExtractText to ToString migration, and merged NameplateNodeService into NameplateService per @nebel's great pointer. FontManager and GlobalScale, and a few stray ImRaii sites I missed will come next, thank you for the patience on my slow work.

MidoriKami:
[Non Blocking Suggestion] You're missing `private` specifiers on a lot of things across the project, while this isn't an issue at all as everything defaults to private, it's somewhat inconsistent across your project.

[Non Blocking Question] I think you have the wrong character listed on line 60 of ChatReplaceService.

Completed review of [ac04d77](https://github.com/rphubco/glance/commit/ac04d776d54d629d3b4fbaf2e0716ae20b19d1e0) and looks good to me 👍 

Your code is rather nice and organized. I would like to see some more consistency with access modifiers, but that's a pretty minor complaint. You can do that at any time.

nathanctech:
FYI @rphubco, someone reported an issue (currently loaded as a custom repo) in case you needed to make a change before this is pushed:

```
System.InvalidOperationException: Operation is not valid due to the current state of the object.
   at Lumina.Excel.RowRef`1.get_Value()
   at Glance.UI.Windows.ToolboxWindow.RefreshActiveProfile() in /_/UI/Windows/ToolboxWindow.cs:line 125
   at Glance.UI.Windows.ToolboxWindow.DrawConditions() in /_/UI/Windows/ToolboxWindow.cs:line 43
   at Dalamud.Interface.Windowing.WindowHost.DrawInternal(WindowDrawParameters internalDrawParams, WindowSystemPersistence persistence) in /_/Dalamud/Interface/Windowing/WindowHost.cs:line 180
   at Dalamud.Interface.Windowing.WindowSystem.Draw() in /_/Dalamud/Interface/Windowing/WindowSystem.cs:line 126
   at Dalamud.Interface.UiBuilder.OnDraw() in /_/Dalamud/Interface/UiBuilder.cs:line 753
```

rphubco:
> FYI @rphubco, someone reported an issue (currently loaded as a custom repo) in case you needed to make a change before this is pushed:
> 
> ```
> System.InvalidOperationException: Operation is not valid due to the current state of the object.
>    at Lumina.Excel.RowRef`1.get_Value()
>    at Glance.UI.Windows.ToolboxWindow.RefreshActiveProfile() in /_/UI/Windows/ToolboxWindow.cs:line 125
>    at Glance.UI.Windows.ToolboxWindow.DrawConditions() in /_/UI/Windows/ToolboxWindow.cs:line 43
>    at Dalamud.Interface.Windowing.WindowHost.DrawInternal(WindowDrawParameters internalDrawParams, WindowSystemPersistence persistence) in /_/Dalamud/Interface/Windowing/WindowHost.cs:line 180
>    at Dalamud.Interface.Windowing.WindowSystem.Draw() in /_/Dalamud/Interface/Windowing/WindowSystem.cs:line 126
>    at Dalamud.Interface.UiBuilder.OnDraw() in /_/Dalamud/Interface/UiBuilder.cs:line 753
> ```

Thanks! RefreshActiveProfile is not handling the title screen so i'll push a fix shortly.
Also caught a nameplate tint leak onto NPCs/mobs while in there.

reiichi001:
Hi @rphubco,

PAC has been discussing the nameplate/chat alias features in Glance, and we'll need you to make a few adjustments to how those work/display.

For the chatbox, please make character aliases additive, so that the original character name is still displayed. A plugin like NickNamer can be used as reference, but how you prefix/append it up to you! 

For nameplates, we do still need the original/actual character name to be displayed in some fashion. There are a couple reasons for this. First, we don't want people to use Glance as a means to cheat SE's paid name change service (one of our core guidelines to ensure we don't get their attention). Second, a plugin user should always be able to find the actual name, whether for `/tell`s, reporting, or other game mechanics. One PAC member suggested that you could use override the nameplate title with a RP character's name instead. (I know that would remove the ability for custom titles). Or you could perhaps generate an imgui overlay when the character's nameplate is hovered to show their in-character details.

If the nameplate has sufficient room to add a third line, or longer top/bottom lines, they can also be extended, but I'm not fully up to date on whether SE has a character limit, and if that can/should be bypassed.

nebel:
Note that it is possible to add an additional "line" to the name part of nameplates just by inserting a `\n` in the name text via the handler. The collision for such plates is actually fine I think. There is the downside that the job icon can appear floating when one name is longer than the other, but it may be acceptable. Although if the plate has two names on it I'd want the formatting/styling to make it clear which name is the real one and which is the fake one.

nathanctech:
Hi @rphubco, just making sure you've seen the comments mentioned and if you're still working on this.

rphubco:
> Hi @rphubco, just making sure you've seen the comments mentioned and if you're still working on this.

Hey, I have and will soon push an update to incorporate these changes plus a few new features in Glance. I have been stuck re-designing the main rphub website, sorry for no follow ups.

nathanctech:
Hi, just to keep things tidy I will be closing this pull request. Please re-open this PR when you are ready to submit for review once more! You will not have to go through the full approval process again.

PR #8096, opened 2026-04-14
MogMogCheck v3.3.0

Haselnussbomber:
- **Added:** The checkmark for equippable items is now also displayed when the item is collected in an Outfit.

### DIFF
--- a/stable/MogMogCheck/manifest.toml
+++ b/stable/MogMogCheck/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/Haselnussbomber/MogMogCheck.git"
-commit = "803e435b281448bf9b38f7aa6dd491d7cf6b1502"
+commit = "4bdacb1b4ffcec1e96932979ff72ddf5f90436ee"
 owners = [ "Haselnussbomber" ]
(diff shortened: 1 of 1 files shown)


Haselnussbomber:
**AI Disclosure:**
I occasionally use Gemini (so, AI on the Assist level) for specific, technical questions, like "How do I do X?" instead of using a search engine. I might copy-paste a code snippet it generates based on this.

For example, in this update HaselCommon contains a new [ImCursor.SameLineSpace](https://github.com/Haselnussbomber/HaselCommon/blob/fe0b266d/HaselCommon/Gui/ImCursor.cs#L34-L41) function, which was mostly generated by Gemini, but this was after I've checked the implementation of [ImGui:.CalcTextSizeA](https://github.com/ocornut/imgui/blob/v1.88/imgui_draw.cpp#L3504), which did way more than I needed.
The [ImRaiiListClipper Enumerator](https://github.com/Haselnussbomber/HaselCommon/blob/fe0b266d/HaselCommon/Gui/ImRaiiListClipper.cs#L49-L66) also was generated by Gemini. I tried to rewrite it but ended up just making sure it works correctly.^^

nathanctech (approved):

PR #8100, opened 2026-04-15
[Testing] KillFeed

wolfcomp:
Just a thing I made as a joke and have had laying around for almost a year now. Might as well try to see if it gets approved or not.

What this does can be seen in the picture here:
image

I full well accept that this can be denied due to the nature of what it does and only want to see if it could or not.


nathanctech:
Hi, I haven't tested this yet but it doesn't look like it disables itself in PvP which would be required. I'll let you know if there are any other concerns as it works through the review queue.

wolfcomp:
Going to wait for the full list of thing with both code and rule review to comment fully about my thoughts, but will note that the setting they added a few patches ago to the game with `Character Display Range` within PvP this is always set at the lowest setting for everyone.

This means that the plugin can only see what you effectivly can see by turning the camera around.

yfthbr:
Code review as of [bbdaa36](https://github.com/wolfcomp/KillFeed/commit/bbdaa36c32fcdd95a0e0522fea1d1038bf204089):
- ImGui.Begin return value not checked in `ActiveNotification.Draw` (should probably use a Dalamud Window instead, though not strictly necessary)

---

> This means that the plugin can only see what you effectivly can see by turning the camera around.

not having to pan the camera around is a significant advantage in my opinion

nathanctech:

wolfcomp:
> Code review as of [bbdaa36](https://github.com/wolfcomp/KillFeed/commit/bbdaa36c32fcdd95a0e0522fea1d1038bf204089):
> 
> * ImGui.Begin return value not checked in `ActiveNotification.Draw` (should probably use a Dalamud Window instead, though not strictly necessary)
> 
> > This means that the plugin can only see what you effectivly can see by turning the camera around.
> 
> not having to pan the camera around is a significant advantage in my opinion

With ImGui and setting the flags 
```csharp
ImGuiWindowFlags.AlwaysAutoResize |
ImGuiWindowFlags.NoDecoration |
ImGuiWindowFlags.NoMove |
ImGuiWindowFlags.NoFocusOnAppearing |
ImGuiWindowFlags.NoDocking |
ImGuiWindowFlags.NoInputs |
ImGuiWindowFlags.NoSavedSettings
```
this forces the window to always be drawn.
```csharp
ImGuiHelpers.ForceNextWindowMainViewport();
ImGui.SetNextWindowPos(topLeft + viewportPos, ImGuiCond.Always, pivot);
var size = new Vector2(width, actionWindowHeight);
ImGui.SetNextWindowSizeConstraints(size, size);
```
Forces the draw to be from the bottom right of the main screen.

yfthbr:
Yeah that seems fine as far as I'm concerned.

nathanctech:
Hi,

Sorry there hasn't been any feedback on this plugin. PAC has decided against allowing this into the main repository as it doesn't seem to be a good fit for the current ecosystem. PAC had also expressed some concerns about how some things were implemented a while ago in the development channels. Thanks.

PR #8116, opened 2026-04-17
VanillaPlus

MidoriKami:
Add Inventory Cooldowns - Shows the cooldown on items in the inventory!


MidoriKami:
Size is mostly from moving functions into their own helper class.

reiichi001 (approved):

PR #8124, opened 2026-04-18
VanillaPlus

MidoriKami:
Update Translations

Prevent BetterSelectString from effecting things that are not SelectString.


nathanctech (approved):

MidoriKami:
No changes to VanillaPlus itself, just localization.

Only change is in KTK

PR #8125, opened 2026-04-18
VanillaPlus

MidoriKami:
Fix Map Overlays not using correct offsets for some maps.

Fix BetterSelectString bug from last update.

More Translations.


MidoriKami:
size-mid due to translations, VanillaPlus code is unchanged, KTK code is changed.

reiichi001 (approved):

PR #8151, opened 2026-04-25
VanillaPlus

MidoriKami:
Add Suppress Strategy Board - Completely suppresses shared strategy boards, no sound effect, no popup!

Remove Fade Strategy Board - Removed in favor of Suppress Strategy Board

Update Translations


MidoriKami:
New strategy board feature is pretty small, the rest of the size is some function refactoring (reordering/new args), and localization.

nathanctech (approved):

PR #8182, opened 2026-04-29
[RetainerSales] dalamud 15 bump

Populo:
AI disclosure in repo readme, copy/pasted here:

Mostly Hint with a splash of some Assist, used as a replacement for google search


Populo:
i switched to a self hosted forgejo from github so that might be causing this error, if it is lmk and ill just push the same update to the github repo that worked before

nathanctech:
That is an odd one, I can clone locally just fine. Maybe IP blocks? It should be coming from Github's range.

nathanctech:
bleatbot, rebuild

nathanctech:
Yeah, not sure why it won't work. Github repo should work at least.

reiichi001 (approved):

PR #8204, opened 2026-04-29
Updating 6 plugins with minimal code changes

Aida-Enna:
What it says on the tin


nathanctech:
Have to update the csproj version of it as well. :) You no longer need a `<PropertyGroup>` for the dependencies since the SDK handles it.

reiichi001 (approved):

PR #8212, opened 2026-04-29
AdaptiveHud 1.0.15.1

Helios747:
goat requested this be resubmitted to fix CI. No code changes, just version bump

PR #8220, opened 2026-04-29
Revert "AdaptiveHud 1.0.15.1"

nathanctech:
Reverts goatcorp/DalamudPluginsD17#8212


Helios747:
ah its fine, probably. I ADHD'd and forgot to push the commit. Is it actually a problem if all the version numbers weren't bumped?

nathanctech:
Should be fine, worst case we'll bump again.

PR #8236, opened 2026-04-30
BeastieBuddy-v7.5.1

rail2025:
api bump +


rail2025:
not sure why its saying lower version...

rail2025:
is this because 7.5 < 7.45 ? 
so i'd have to either go 7.50 ? or you could just be like elsa?

nathanctech:
Yes, it's considering 7.45 to be < 7.5 as it just checks the individual numbers, so you'd need it to be 7.50 or so.

reiichi001 (approved):

PR #8268, opened 2026-04-30
[TermFilter2] A rewrite of the Term Filter feature with extreme customization

Aida-Enna:
# Term Filter 2

An expanded and extremely customizable rewrite of the in-game term filter, this allows players to mask/hide/modify messages containing specific words. I felt like the vanilla term filter was missing a lot of options, so I added them.

https://github.com/user-attachments/assets/29cb7022-e5ff-448d-b4bb-e6c56e53fe88

With each filter (there is no limit to how many you can have), you can customize each of the following:

* Which chat channel(s) you want it to filter, or all of them!
* Which player(s) you want it to filter, or all of them!
* Whether to hide the message completely (and log it in the /xllog) or show a red "[Filtered] Message could not be displayed due to Term Filter 1."
* Whether to (instead of hiding) replace *only the phrase* in text with another of your choice (or multiple ones it'll pick at random) and bold it so you know it's been replaced.

"If the term filter is so good, why didn't they make a term filter 2?"


Aida-Enna:
Ahhh, a SLN issue from when it was standalone, will fix

foophoof (commented):
Code reviewed https://github.com/Aida-Enna/TermFilter2/commit/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76.

Code mostly looks OK to me, and plugin loads fine (although I got an exception the first time I opened the main window, see below).

**note (non-blocking):** I wasn't able to go as deep into looking for bugs as I normally try to do, the `Plugin.ChatMessage` and `MainWindow.Draw` methods are quite long and not super easy for me to fully reason about. I'd encourage you to try to split those up into smaller methods, I think they'd be easier to maintain and understand going forward.

---

Two blocking things that should be fixed:

**todo (blocking):** Missing plugin introduction in Discord message.

As this is the initial release of a new plugin, please either update the changelog to be an introduction for the plugin, or remove the changelog from the TOML file and make the PR description be an introduction for the plugin, since this is what gets posted to Discord (right now it would just say "Initial release").

**issue (blocking):** Leaked action subscription in `Plugin`

The Plugin constructor subscribes to `ClientState.Login`, but there's no corresponding unsubscription in the Dispose method.

---

Also, some non-blocking suggestions/notes. Feel free to use these if you like, but you can also ignore all of them if you want:

**issue (non-blocking):** I got an exception when opening the main window after installing the plugin:

```
System.Collections.Generic.KeyNotFoundException: The given key '[redacted]' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at TermFilter2.Windows.MainWindow.Draw() in /home/sarah/code/TermFilter2/Windows/MainWindow.cs:line 78
   at Dalamud.Interface.Windowing.WindowHost.DrawInternal(WindowDrawParameters internalDrawParams, WindowSystemPersistence persistence) in /_/Dalamud/Interface/Windowing/WindowHost.cs:line 342
```

I suspect this is because I installed the plugin before I logged in, so my content ID wasn't added to the configuration. If I click "Attempt to retry" it loads fine, so probably just a one-time setup thing.

I also noticed that a "0" content ID got added to the config:

```json
{
  "$type": "TermFilter2.Configuration, TermFilter2",
  "Version": 0,
  "Terms": {
    "$type": "System.Collections.Generic.Dictionary`2[[System.UInt64, System.Private.CoreLib],[TermFilter2.TermFilter2Collection, TermFilter2]], System.Private.CoreLib",
    "0": [],
    "[redacted]": []
  }
}
```

I suspect this is because of

```csharp
if (!PluginConfig.Terms.ContainsKey(PlayerState.ContentId))
{
    PluginConfig.Terms.Add(PlayerState.ContentId, new TermFilter2Collection());
    PluginConfig.Save();
}
```

`PlayerState.ContentId` is 0 when the PlayerState isn't loaded, it might be worth considering `PlayerState.IsLoaded && !PluginConfig.Terms.ContainsKey(PlayerState.ContentId)`?

**todo (non-blocking):** There's a few things that seem to be leftovers from older plugin templates, but that are no longer used, so they can be cleaned up:

- It's no longer necessary to set [`Name`](https://github.com/Aida-Enna/TermFilter2/blob/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76/Plugin.cs#L22) in the `IDalamudPlugin` implementation.
- The [`GameGui`](https://github.com/Aida-Enna/TermFilter2/blob/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76/Plugin.cs#L27) and [`ContextMenu`](https://github.com/Aida-Enna/TermFilter2/blob/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76/Plugin.cs#L29) services are unused in the plugin, so those attributes can be removed.
- It's not necessary to set [`TargetFramework`](https://github.com/Aida-Enna/TermFilter2/blob/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76/TermFilter2.csproj#L4) in the csproj, this is set automatically by Dalamud.NET.Sdk (I also don't think the `OutputPath` and `ItemGroup` attributes are needed, unless you have a specific reason to override those).
- [`"ApplicableVersion": "any"`](https://github.com/Aida-Enna/TermFilter2/blob/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76/TermFilter2.json#L4) should be removed from TermFilter2.json, this is set automatically by the build automation.
- The [plugin constructor](https://github.com/Aida-Enna/TermFilter2/blob/b265a6b07a3c10c4bdbbbf27bdc3fa17ac78ce76/Plugin.cs#L41) accepts a lot of unused plugin  (the only one that's uses in the constructor is the `ICommandManager`, it might be worth moving that to be a `[PluginService]` attribute for consistency.).

**question (non-blocking):** It looks like messages from the local player is detected (and skipped for processing) like this:
```csharp
if ((Sender.Payloads.Where(a => a.Type == PayloadType.RawText).Any(x => ((TextPayload)x).Text == PlayerState.CharacterName))) { return; }
```
I'm not all that familiar with `IChatMessage`, but from the docs it seems like `Sender` can be modified by other plugins, if there are multiple plugins subscribing to the `ChatMessage` event. It might be more robust to check `message.SourceKind == XivChatRelationKind.LocalPlayer`, or checking the name against `OriginalSender` instead.

**question (non-blocking):** It looks like the player name in the message has all non-whitespace and non-alphanumeric characters stripped before comparing against the saved name list. Player names can contain some other characters as well, like `-` and `'`, would those names be handled correctly?

**suggestion (non-blocking)**:  I'd strongly encourage using ImRaii instead of ImGui Push/Pop and Begin/End methods. For example
```csharp
private void DrawTermsTable()
{
    using var table = ImRaii.Table("##TermsTable", 8, ImGuiTableFlags.Borders | ImGuiTableFlags.SizingFixedFit);
    if (!table.Success)
        return;

    ImGui.TableSetupColumn("Term");
    // […]
}
```
and
```csharp
using (ImRaii.PushColor(ImGuiCol.Button, new System.Numerics.Vector4(0.25f, 0.67f, 0.87f, 1)))
{
    if (ImGui.Button("Support me on Ko-Fi"))
    {
        Dalamud.Utility.Util.OpenLink("https://ko-fi.com/Y8Y114PMT");
    }
}
```

Aida-Enna:
Thanks for the lengthy (in a positive way!) review - It helped me learn a few things I didn't know and I genuinely appreciate it!

I've made most of the recommended changes, but for ``OutputPath`` I do override it because of my testing environment.

I'll also seperate those methods into smaller ones and switch to ImRaii in a (soon) future update 👍

foophoof (commented):
Reviewed https://github.com/Aida-Enna/TermFilter2/commit/31ff5adb0d8be278ace72342820271dbf59caea9, LGTM 👍🏻

nathanctech:
Tested, loads/unloads fine with no errors. Thanks for your contribution!

nathanctech (approved):

PR #8269, opened 2026-04-30
Update Orange Guidance Tomestone to 1.10.6

anna-is-cute:
API 15

### DIFF
--- a/stable/OrangeGuidanceTomestone/manifest.toml
+++ b/stable/OrangeGuidanceTomestone/manifest.toml
@@ -1,9 +1,8 @@
 [plugin]
 repository = 'https://git.anna.lgbt/anna/OrangeGuidanceTomestone.git'
-commit = '9a3445a614d0b591b9ad8bd8e4870b1246c66cf6'
+commit = 'ae9dcc8b0a91504746159494434872bf6eb61db2'
 owners = [ 'anna-is-cute' ]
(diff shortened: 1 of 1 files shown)


reiichi001:
Since the last time this plugin was updated, we've received a handful (not constant, but somewhat consistently a few each month) of user feedback posts related to being unable to reach the OGT server. (Also ExtraChat, but I was gonna wait to mention that in its own PR)

Is there somewhere we can direct users for support (or really, "is it down or just me") questions? Or would you be willing to set `AcceptsFeedback` to false on the plugin to disable the button?

reiichi001 (approved):

anna-is-cute:
Not like I can see the feedback anyway, so accepts feedback being turned off is fine. It's weird that OGT would go down for anyone; more likely weird routing shit or CF blips. ExtraChat does sometimes deadlock and need a restart.

Should I just submit another PR here?

reiichi001:
A fresh PR works. When I've grabbed logs, it's sometimes a TLS handling issue, so it could be CF or a routing issue that's regional or specific to the user. When we do get reports about it, it's usually consistent for a period of time for the user, and then it "suddenly" starts or stops working, according to them. The hard part is that it's not widespread enough to remember to save a log for communication purposes like this, but it's just common enough that it generates chatter on the discord.

PR #8270, opened 2026-04-30
Update XIVChat to 1.7.13

anna-is-cute:
API 15


anna-is-cute:
Forgot I wasn't using jj so a file was missing, oops.

reiichi001 (approved):

PR #8279, opened 2026-04-30
Currency Watchdog 1.0.2.3

nebel:
- Update for API 15
- Fix an issue where a burden's advanced limit settings would be cleared when using Revert or Clone.


nebel:
Meant to do stable!

PR #8303, opened 2026-04-30
CrossUp 1.7.1.3

ItsBexy:
Updated for 7.5 / API 15


nathanctech:
Please remove the DalamudPackager reference from your csproj as it could cause issues in the future.

MidoriKami:
Additionally Line 12 of CrossUp.yml is another API version override, this shouldn't be set at all and should removed from the yml file

reiichi001 (approved):

PR #8320, opened 2026-04-30
VanillaHotbarExtender 1.2.6

mustafakalash:
I switched over to the Dalamud SDK but didn't install the package from nuget, so while the dev build loaded fine, it was giving an error when updating/enabling the main repo version.

### DIFF
--- a/testing/live/VanillaHotbarExtender/manifest.toml
+++ b/testing/live/VanillaHotbarExtender/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/mustafakalash/VanillaHotbarExtender.git"
-commit = "fb24dd11e99a99fdf3012918e77f0b8c07598351"
+commit = "3074371003f1c5b6a69027b387327394aca14709"
 owners = ["mustafakalash"]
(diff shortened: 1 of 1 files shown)


nathanctech (changes requested):
Hmm, typically just using the SDK should be fine - are you using Visual Studio 2026? I'd like to avoid the Nuget style of including the dalamud stuff if possible. If you can provide the error(s) you get when using the standard SDK that would be helpful.

mustafakalash:
Ah okay, thanks for the info. On further inspection, it seems that the updater did install the latest version and I can enable the main repo plugin and use it normally. However, it is stuck showing that there's an update available and gives this pop up:
image
The logs just show `INF | [PluginManager] Plugin "Vanilla Hotbar Extender" is already up to date`. I'm not sure if adding the SDK as a package made any difference, since loading the dll manually is working fine regardless.

It's even listing both the current and available version as the same:
image
image

nathanctech:
There's currently a known bug with testing exclusive plugins that's being looked into, nothing you've done.

mustafakalash:
Ohh, got it. Thank you!

PR #8371, opened 2026-05-01
SpotifyHonorific: Initial submission

Valiice:
SpotifyHonorific displays your currently playing Spotify track as your FFXIV honorific title via the Honorific plugin's IPC.
	- Requires the https://github.com/Caraxi/Honorific plugin
	- Authenticates via Spotify OAuth (PKCE flow, no credentials stored)
 	- Supports Scriban templates for customizable title formatting
 	- Pauses polling when AFK (30s idle threshold, Windows only - gracefully disabled on Linux)
 	- Multiple activity profiles with per-profile templates
 
AI Disclosure: Assist


nathanctech:
Hi, please place your manifest in the `testing/live` folder instead of `stable`.

Valiice:
Will do, whats the average time that it takes for it to be moved from testing to stable? Since it was a plugin that was available via my custom repo before

nathanctech:
Typically that's up to the dev, but all new plugins have to start in the testing branch for their initial release. Your plugin is in our review queue now.

Valiice:
Bumped commit to 6e24f85 (1.12.0.0): adds auth warning toast + dep bumps.

nathanctech:
Reviewed commit: https://github.com/Valiice/SpotifyHonorific/commit/6e24f8546112172514b0028f01bf4209260d926f

[Blocking] Remove from csproj as this is bundled within the SDK:

```
    <ItemGroup>
      <PackageReference Update="DalamudPackager" Version="15.0.0" />
      <PackageReference Update="DotNet.ReproducibleBuilds" Version="2.0.2" />
    </ItemGroup>
```

[Blocking] You require a client secret to be set for Spotify, but PKCE flow does not require this and you subsequently do not use this in your code. Please remove it.

[Blocking] You dispose the SpotifyAuthenticator when the config window is closed, but it's also being disposed on plugin unload which will throw an error. Add a IsDisposed guard or remove one of the Dispose if you don't require it in both.

[Blocking] HandleAfkStatus() should be looking at the AFK status of the user (or just use the framework timer) rather than using Win32-specific APIs which would make this feature incompatible with Linux/Mac users. The game has a built in AFK state if you want this, or use the game's timestamps.

[Non-blocking] I highly recommend using [ImRaii](https://dalamud.dev/api/Dalamud.Interface.Utility.Raii/Classes/ImRaii/) where possible as it's much safer to use, especially in your Draw() function.

As for compliance, I'm wondering what the intention of this plugin is since local-only Honorific title changing seems not useful to me, and therefore this plugin would likely be used in conjunction with custom repo plugins and would then run against our rules of having custom repo plugin dependencies. I'm curious as to your thoughts on this.

Valiice:
I've reviewed and fixed all four blockers, and went ahead with the ImRaii cleanup too since it's safer and cleaner.

On the compliance question, SpotifyHonorific is intended as a local UX feature. It's essentially a personal track viewer surfaced inside FFXIV, the same kind of local "now playing" readout your Spotify desktop window or taskbar mini player already shows you, only visible to you. The title renders on the user's own nameplate. That's the same personal UI surface as a lot of accepted local only plugins, and it's enjoyable as ambient feedback for the user themselves without anyone else needing to see it. The code itself has no references to, IPC into, or checks for any custom repo plugin. The only IPC contract used is Honorific's own (Honorific.SetCharacterTitle and Honorific.ClearCharacterTitle), which is from an official repo plugin.

nathanctech:
While true it's good as a now playing type of thing, most players won't have their own nameplates enabled to see it as it takes up some screen real estate. Thanks for updating the code - I will review and keep you posted.

nathanctech:
bleatbot, rebuild

nathanctech:
Hello,

Upon a review with other members of PAC, we have made the decision to not allow this plugin into the main repository due to its limited scope of use (unless third party repos are used) and therefore would not be a good fit for the ecosystem as it stands. PAC also had concerns of it being a clone of an existing plugin with heavy LLM assistance to make it work with Spotify instead. Thank you for your contribution and you are free to distribute it as a custom repository plugin.

PR #8416, opened 2026-05-01
Update code to support 7.5

jlkeesey:


nathanctech (changes requested):
Please remove `<PackageReference Update="DalamudPackager" Version="15.0.0" />` from your csproj as this is no longer supported.

nathanctech (approved):

PR #8425, opened 2026-05-02
[Testing] Discord Rich Presence 2.0.9.0

Bronya-Rand:
## What's New?
Rewrote all the code of the RPC plugin to use a template over a bunch of special cases. Now all you need to do is follow a template of tags and assign it how you want and boom, you're off to the races.

This PR is not intended to be backwards compatible with the old configuration file and a migration is provided to move to V2. Do note if all is good here, this is the new RPC system.


reiichi001:
Actual diff: https://github.com/Bronya-Rand/Dalamud.RichPresence/compare/9d4d55ab5e2240e5b04ebbdca0065aff71ee4489...e4dd60da76dfae2eddc4d72036f6c307d75bbedb

nathanctech (approved):

PR #8427, opened 2026-05-02
Add QuickSymbols plugin

LatencyBryer:
This plugin is intentionally small and focused on just leting players being able to insert special chat symbols into some in-game text fields such as `Party Finder`, `House Guest Book` and `Chatlog`. Symbol insertion only happens through native input insertion when the intended input is active and the plugin does not hook chat send flow.

### Overlay button system

- Adds a custom `♥` overlay button anchored near the native ChatLog chat mode button
- Button size is derived from the native ChatLog UI placement so it follows the player's UI scale and ChatLog layout
- Supports a manually adjusted button position using a saved relative offset from the native ChatLog position
- The saved position follows ChatLog movement instead of being stored as a fully fixed screen coordinate

### Detection system

- Uses `IGameGui.GetAddonByName(...)` to locate native game UI addons
- Scans both regular node traversal and `UldManager.NodeList` for text input components where needed
- Adds `CTRL + Click` support for favoriting and unfavoriting symbols in the popup list
  - _Favorite symbols are stored in plugin configuration and maintain saved through game sessions so it doesnt reset after logout/crash and etc_

### UI Theme Adaptation

- Reads the game's current UI theme using `IGameConfig` and `ColorThemeType`
- Applies plugin colors based on the player's selected game UI theme
- Supports visual mappings for all available game UI theme options

> **Dalamud theme is currently ignored and given priority to use the detected in-game theme applyed to the UI since the idea isn't to make it look like a plugin window but a native in-game button instead.**


seventity7:
sorry forgot to send this here too

AI assistance was used at the **“Assist”** level for the plugin code
[AI-DECLARATION.md](https://github.com/user-attachments/files/27296241/AI-DECLARATION.md)

seventity7:
Plugin updated May 02/2026

I was afraid that the method I used for UI theme would cause leaks to other Dalamud plugin Windows so I pushed a fix to prevent that from happening.
  - I replaced `ImGui.PushStyleVar`, `ImGui.PushStyleColor`, `PopStyleVar` and `PopStyleColor` with `ImRaii.PushStyle(...)` and `ImRaii.PushColor(...)` using `using var` / `using (...)`. This automatically disposes of the scope and significantly reduces the risk of visual leakage to other Dalamud windows.

seventity7:
Plugin updated May 03/2026

- Added `PluginConfiguration.Custom` to persist user-defined custom entries
- Added keybind handling in `FrameworkUpdate()` to open the popup near the current mouse position
- Updated `DrawSymbolsPopup(...)` to render tabbed content instead of only the built-in symbol grid
- Updated insertion handling so it now also supports any currently focused text input instead of just `ChatLog` `Party Finder` `House Guest Book`

testing/live/QuickSymbols/manifest.toml
```diff
@@ -0,0 +1,7 @@
+[plugin]
+repository = "https://github.com/seventity7/QuickSymbols.git"
+commit = "a07a02f799cb79d654a49cecdd5c0659f8c1e95e"
+owners = ["seventity7"]
+maintainers = ["seventity7"]
+project_path = "."
+changelog = "Update 1.0.0.2 | new features"
```
yfthbr: the first public release should come with a changelog that briefly introduces the plugin and what it allows users to do.

yfthbr (commented):
Partial review as of [a07a02f](https://github.com/seventity7/QuickSymbols/commit/a07a02f799cb79d654a49cecdd5c0659f8c1e95e)

1. Your OpenMainWindow also opens the config window. It's better to leave OpenMainUi un-registered than have two buttons that do the same thing.
2. `GetFocusedTextInput` only the first way of finding the input element checks that it is enabled before using it, but it looks like the fallback after the loop can also use the same node without the enabled check? Is that correct?
3. `CheckHotkeyState` loops over all virtualkeys and it's not entirely clear from reading it as to why it has to do that.
4. using ImGui IO to configure hotkeys but KeyState to check them is a little odd, and there could be a mismatch between the 0-159 used to configure the keys and KeyState's valid virtual keys used in `CheckHotkeyState`. It would be good to use KeyState (or IO, if there's a specific reason for it) for both.
5. Consider dropping DalamudApiLevel from the QuickSymbols.json manifest, as the SDK should handle it for you already, unless you have a specific reason for keeping it there too.

Reviewed thus far:
- plugin manifest
- ctor, dtor
- FrameworkUpdate flow
- UiBuilder Draw flow in execution order _up to but excluding_ `TryGetRecruitmentCommentTarget` call in Draw()

LatencyBryer:
Hi!
I will remove the `OpenMainUi` registration instead of keeping two Dalamud buttons that perform the same action.

**`GetFocusedTextInput()` enabled check:**

The first path checks `input->Enabled` while walking up the focused node hierarchy but the fallback path can return the parent component as an `AtkComponentTextInput*` without checking whether that text input is enabled or not.
That was not intentional. I will update the fallback so it only returns the text input when the component type is `TextInput` **and** the resulting `AtkComponentTextInput` is enabled. Otherwise it will return `null`.

**`CheckHotkeyState()` looping through all virtual keys:**

The intent was to require an exact hotkey match and avoid triggering when extra keys are held but I agree that this is not very clear from the code and is probably stricter than necessary.
I will simplify this so the check only verifies that all configured keys are currently pressed and then uses the existing `hotkeyWasDown` guard to avoid repeating while the same combo is held.

**Hotkey configuration using ImGui IO while hotkey checking uses `KeyState`:**

I will change the hotkey recording path to use `KeyState.GetValidVirtualKeys()` as well, so recording and runtime checking use the same source and the stored values are valid for `KeyState`.

**`DalamudApiLevel` in `QuickSymbols.json`:**

I do not have a specific reason to keep both `DalamudApiLevel` `Dalamud.NET.Sdk/15.0.0`, So I will remove the `DalamudApiLevel` entry and let the SDK handle it.
I'll update the PR with these changes soon!
Thanks again.

LatencyBryer:
- Removed `DalamudApiLevel` from manifest so the SDK can handle the API level.
- Removed `InformationalVersion` `FileVersion` `AssemblyVersion` from `.csproj` since SDK can handle it from `Version` alone
- `OpenMainUi` no longer opens the config window.
> But I kept a main UI callback registered so the plugin does not trigger Dalamud `the plugin does not register a main UI callback` warning but it now has separate behavior from `OpenConfigUi`
  - `OpenConfigUi` opens the Quick Symbols config window
  - `OpenMainUi` opens the symbol picker UI
    - If a text input is currently focused, it opens the contextual picker near the mouse for that input otherwise it opens the normal Quick Symbols picker

`GetFocusedTextInput()` now checks `Enabled` in both paths.
The fallback path now also validates the resolved `AtkComponentTextInput` before returning it:
```csharp
var inputComponent = (AtkComponentTextInput*)focusParentComponent;
return inputComponent->Enabled ? inputComponent : null;
```

`CheckHotkeyState()` no longer loops over every valid virtual key to decide whether the configured hotkey is pressed.
It now only checks the configured keys and returns once the full configured chord is pressed. This should avoid unrelated pressed keys affecting the result.

The hotkey configuration path no longer uses ImGui IO key arrays. Now the plugin uses `KeyState.GetValidVirtualKeys()` for both:
  - capturing/configuring the hotkey in the config UI;
  - checking the configured hotkey during framework updates
> _This removes the previous mismatch risk between ImGui's `0..159` key range and Dalamud `KeyState` virtual keys_

Configured hotkeys are normalized through the same valid `KeyState` virtual-key set. This also filters out mouse buttons and keeps saved hotkeys restricted to keys that `KeyState` can actually validate later.

### Important
While testing the hotkey changes, I also adjusted the keybind flow so it behaves more **safely** in-game:
- To avoid focus issues while recording a new hotkey, the keybind display no longer uses a read-only `InputText`
- `Set Keybind` starts recording cleanly on the next frame;
  - `Esc` cancels recording;
- while recording a new hotkey, pressed keys are cleared through `KeyState` to reduce the chance of the game handling the same chord at the same time;
- when using a configured modifier + text-key hotkey such as `Alt+S` for example, with a focused text input, the plugin attempts to prevent the text key from being left in the input before the symbol picker UI opens.

_(btw just saw the note on the manifest about changelog after submiting it, I'll do another before release if needed)_

yfthbr:
Fully reviewed as of [d8cc084](https://github.com/seventity7/QuickSymbols/commit/d8cc0849f25b252c33d27b132f091f7eb67b74d0)

- **(Important)** `DrawEntriesGrid` calls `EndChild` even when `BeginChild` returned false. ImGui elements must only be ended when they succeeded. The only exception to this is the top level `ImGui.Begin` that creates a new window. You can change the function to use `ImRaii.Child` instead. (The `IDisposable? pushedFont` can also use `using`, which can happily handle `null` too, just `using (pushedFont) { ... }` around the for-loop)
- (Potential bug) `PickBestRecruitmentCommentCandidate` and `PickBestMessageBookInputCandidate` use ImGuiHelpers.GlobalScale, but you probably shouldn't use that there, as the functions concern game UI, not ImGui, and the scaling can be completely different. You likely want to use each addon scale, if the size doesn't already take that into account (I'm not sure, you'd have to check)

Likely doesn't need changing, but `SendKeyPress` is checking `OperatingSystem.IsWindows()`, which will always be true as Dalamud only currently always runs the Windows version of the .NET runtime. _If_ Linux users with wine are a concern, you should use `Dalamud.Utility.Util.IsWine()` instead, but I _suspect it is not required_ for your use-case.

LatencyBryer:
> bleatbot error doesnt seen to come from my end

Fixed the DrawEntriesGrid child handling by switching it to ImRaii.Child so the child scope is only used when BeginChild succeeds. Also wrapped the pushed symbol font in a using block around the grid draw loop.

For the addon candidate picking, I changed the minimum size checks to use the current addon scale instead of ImGuiHelpers.GlobalScale,

I also removed the OperatingSystem.IsWindows() check from SendKeyPress,

yfthbr:
Code looks good [ca3a8e7](https://github.com/seventity7/QuickSymbols/commit/ca3a8e7a6125d9f600680de84591e05107bfaea1)

nathanctech:
Tested in-game, I can't move the symbols from the bottom left of the game window so it's covering my chat window (change button position doesn't seem to do anything?) and the hotkey doesn't work either.

LatencyBryer:
> Tested in-game, I can't move the symbols from the bottom left of the game window so it's covering my chat window (change button position doesn't seem to do anything?) and the hotkey doesn't work either.

Hi! To change the button position first click on the button to open the popup itself and then hit the button `Change button position`, once the button `♥` becomes Red, Drag and Drop it wherever you did like. Once you done positioning it, hit the button `Change button position` again, it should turn the color back to normal.
image

Note that this "move" functionality is only for the button that becomes available with the chat, it wont be available elsewhere.

To use a keybind you first need to setup one. Once in the settings window of the plugin, click specificaly the button `Set Keybind` and hit the Keys combo you want to save. Just remember that, if you press the Keybind without having any kind of input text field in focus, nothing will show up. _(Also, text fields from other plugin windows might not count as one yet)_
> You need to have a text field focused to be able to use the keybind, example: Click the game chat and hit the keybind or go to the Macros page and go to the field where you type stuff for the macro and hit the keybind.
image
 Note that clicking the Keybind field itself wont do nothing since is just display and not a actual button.
 
 Tested everything said above in my game and it worked fine. If possible, tell me if you are Linux/Windows user, what language you using to type, What keys you are trying to save as keybind and a screenshot of the game chat field showing the current button position. 
 
 Thanks

nathanctech:
Ah, the button was partly off my screen initially so I didn't see it, it works now, just a different way of moving the window, heh.

image

I tried this and several other keybinds and they don't seem to work (Windows, English).

Edit: Oh, just re-read your comment about it being text input only...that works, although it displays a second window if the other is still open. No big deal though, you may want some help text there saying that's the behavior since I wouldn't have known.

LatencyBryer:
## Changes 1.0.0.5

### Popup Behavior
- Added a new **`Close popup on lost focus`** setting
  - Disabled by default
  - When enabled, the symbol popup closes after clicking outside of the popup area
    - The option includes a tooltip explaining the behavior
- Now only one symbol popup can be active at a time
  - Prevents duplicating/overlapping symbol popups
- Popups opened through the saved keybind can be moved by dragging empty space inside the popup
  - This temporary position is not saved and resets the next time the keybind popup is opened
- Improved popup/button layering so the `[♥]` button no longer renders above the keybind popup

### Popup Controls
- Replaced the previous reposition text button with a icon/symbol
  - Now shows a tooltip explaining the normal and active repositioning states
- Added a new config shortcut icon to symbol popups
  - Opens the Quick Symbols configuration window directly from the popup

### Configuration Window
- Displays a tooltip explaining how to correctly record a keybind
- Updated the saved keybind display field to look less like a button
- Improved initial configuration window sizing to show all elements, avoiding partial cut-offs

nathanctech (approved):
Review of [eb56fbe](https://github.com/seventity7/QuickSymbols/commit/eb56fbee63f5f6fd79498fdfd760d968f7262991) tested fine. Thanks for your contribution!

PR #8455, opened 2026-05-02
Currency Watchdog 1.0.3.0

nebel:
- Add jurisdictions
  - Jurisdictions can be created in the "Advanced" drop-down header when selecting a burden
  - Jurisdictions can be used to override overlay panel visibility settings when in certain zones or when doing certain types of content


nebel:
Apologies for the settings window refactoring. `BurdensTab.cs` was becoming huge so I split it into multiple parts which made the diff pretty bad.

reiichi001 (approved):

PR #8499, opened 2026-05-03
VanillaPlus

MidoriKami:
nofranz

There are **no** code changes, this is a refactor after applying editorconfig `trim_trailing_whitespace = true`

I suggest using the Semantic diff instead of regular diff.


MidoriKami:
Oh I guess this does remove the ClearTextInputs feature as well, but nothing added 👍

nebel (approved):

PR #8521, opened 2026-05-04
Update "Flash on Ready Check" for 7.5

Pistachio-dev:


reiichi001:
@Pistachio-dev please bump the version of the plugin.

There is also no file named `NewestRelease.json` in the repository.

nathanctech (approved):

PR #8555, opened 2026-05-07
PvP Tracker new owner

wrath16:
Added @rennalol as a project owner. Same username on Discord.


nathanctech:
Plogon doesn't like no-op commits (i.e, no changes). To make things easier, do you mind just doing a version bump? Otherwise, you can change the owner when an update is actually required.

nathanctech (approved):

PR #8565, opened 2026-05-07
[New Plugin] Mannerisms v1.0.0

mustbetommy:
**Plugin: Mannerisms**

Gesture suggestion plugin. Matches what you type in chat against configurable patterns and surfaces emote suggestions in a floating window. Player clicks or uses a keybind to accept (no emotes are automated).

Supports pre-made common gestures, a simple word-based builder, and full regex for advanced users. Per-character profiles, target/party awareness, configurable timeouts and keybinds.

> **AI usage disclosure: Code Completion in-IDE**
> AI assisted with code-completion during implementation (ImGui, C# syntax). Architecture and design decisions are mine.


nathanctech:
bleatbot, approve

Infiziert90:
Gave your plugin a quick look and found number of issues related to ImGui.

- Please do not `GlobalScale` the `SizeContraints`, I'm not even sure what your intention is with this Maximum, just set it to `float.MaxValue`
- Any other hardcoded value of width or height should be `* GlobalScale`
- There is a lot of ImGui `Begin` that isn't if() checked, this should definitely be changed for safety
- I also found a case where `End`is called multiple times, this is a primary crash reason
- ImRaii should really be used with all these Begin/Ends in the code, gives another layer of crash and corruption resistance

MustBeTj:
> Gave your plugin a quick look and found number of issues related to ImGui.
> 
> * Please do not `GlobalScale` the `SizeContraints`, I'm not even sure what your intention is with this Maximum, just set it to `float.MaxValue`
> * Any other hardcoded value of width or height should be `* GlobalScale`
> * There is a lot of ImGui `Begin` that isn't if() checked, this should definitely be changed for safety
> * I also found a case where `End`is called multiple times, this is a primary crash reason
> * ImRaii should really be used with all these Begin/Ends in the code, gives another layer of crash and corruption resistance

Thank you for this! This is my first plugin so I'm still learning the ropes around ImGui. I cleaned things up.

yfthbr:
ECommons is a big dependency to include for the few simple things you're doing here. I'm not saying to avoid using it, but it would simplify the review and help your dependency management going forwards.

Instead of using `ECommons.DalamudServices` you can make your own Svc.cs. Example [definition](https://github.com/yfthbr/musical-guide/blob/aa6d1165ce23eea487818602712ca90cb0c5fbe6/MusicalGuide/S.cs), [initialization](https://github.com/yfthbr/musical-guide/blob/aa6d1165ce23eea487818602712ca90cb0c5fbe6/MusicalGuide/MusicalGuide.cs#L25)

Instead of using `IsNullOrEmpty` from ECommons, use `string.IsNullOrEmpty` directly

For ImGui, textures and fonts look into `ImGuiHelpers`, font and texture management in Dalamud itself; it's all equally or almost equally simple to use.

PR #8566, opened 2026-05-07
Update plugin 'Clock'

LatencyBryer:
### Update May 07 2026

- Removed hard-coded ImGui style overrides so the plugin now follows Dalamud themes
- Replaced chat-message parsing with a Lodestone/RSS-based check
  - Added `LodestoneMaintenanceService` to fetch maintenance posts and parse date/time block
- Replaced the previous hard-coded timezone enum flow with arbitrary `TimeZoneInfo` IDs
- Added legacy enum migration so existing user configurations continue to resolve to the correct timezone
- Added `TimeZoneHelper` for timezone lookup, display labels, abbreviations, country-based lookup and common timezone highlighting
  - Added country-based timezone lookup with a searchable country dropdown list
  - Highlighted common game-related timezones in gold and added a `Commonly Used` tooltip for them
- Labels use timezone abbreviations such as `EST`, `PST`, `GMT` and `MDT` instead of repetitive UTC offset prefixes _(so it looks cleaner)_
- Added support to favorite timezones with a new `Favorite Timezones` section on the tab `General`
  - Favorite timezones are sorted first in the dropdown list and can be clicked directly to switch the active timezone to them
- Added a new `Plugin Config` tab on main window
  - Added `ConfigurationFileService` for full plugin configuration export/import system
  - Added `Export`/`Import` actions for saving/loading all current plugin settings in a configuration file
- Moved non-configuration stuff out of `Configuration.cs`
  - Added dedicated state/service files `AlarmConfigurationState`, `AlarmConfigurationService` and `ConfigurationFeatureState` so `Configuration.cs` is more focused on configuration stuff

### DIFF
--- a/testing/live/Clock/manifest.toml
+++ b/testing/live/Clock/manifest.toml
@@ -1,7 +1,7 @@
 [plugin]
 repository = "https://github.com/seventity7/Clock.git"
-commit = "5493a50d8c171d33aeb278490addeb81f6d261e2"
+commit = "9eb150eaa225a48d4b7d83ed06138d58ee093e5a"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
-changelog = "Oficial Release > Update 1.0.6.4"
+changelog = "Oficial Release > Update by review feedback"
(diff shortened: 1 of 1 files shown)


LatencyBryer:
Updated based on some review feedbacks.

Changes made:
- Replaced the manual tab flag cast with `ImGuiTabItemFlags.Trailing`
- Removed the custom Lodestone User-Agent
- Reduced automatic Lodestone checks to a longer interval
- Changed maintenance lookup to fetch at most one matching maintenance article
- Added caching to avoid repeatedly processing the same maintenance post

nathanctech (approved):

PR #8594, opened 2026-05-09
Update SimpleTriggers v1.0.2.4

Brolijah:
## SimpleTriggers v1.0.2.4
* Minor improvements.
* Disabled dectalk (was in testing)
* Reduced artifact size ~7 MB. (Removed linux binaries from the output build, they aren't needed in Wine)


Brolijah:
DecTalk is in the source code but disabled in release builds. It has too many problems with Memory mode in the library. I may revisit in the future but it's kinda too broken for public use. It *technically* works, but it's a dice roll whenever it initializes if it will work. Unloading and reloading the library also crashes in Windows, also randomly.

nathanctech (approved):

PR #8599, opened 2026-05-09
TargetLines 10.0.1.0

Drahsid:
Fix the crash and error that was reported, one of which was caused by game object references going stale mid-frame; the other was from the faulty assumption that a game object of kind PC also had a PC game object.
There are also various other safety-related changes to ensure other similar issues don't crop up. Also a bugfix related to the target switching animation being skipped for client-only objects.


Drahsid:
Made another minor change that fixed an issue I almost forgot about. It took a while to replicate, but the fallback line was able to overwrite the filter which passed which would cause line colors to change or for them to just go invisible.

nathanctech (approved):

PR #8638, opened 2026-05-12
Update Echoglossian to v4.2601.0512.x

lokinmodar:
## Summary
- update Echoglossian to v4.2601.0512.x
- point the manifest at commit 3ff33f753910d06043ebb389a57c044d73385a49
- correct the submitted version scheme to the new 4.yy01.mmdd.HHmm format


nathanctech (changes requested):
Plugins should not have dynamic versioning at all. Please remove this logic as reproducible builds (same commit hash = same version) is a requirement in D17.

lokinmodar:
> Plugins should not have dynamic versioning at all. Please remove this logic as reproducible builds (same commit hash = same version) is a requirement in D17.

done.

nathanctech (approved):

PR #8642, opened 2026-05-12
AetherBags 1.0.0.4

Zeffuro:
# 1.0.0.4
- Unified Category System is now the that default system.
- Added Crystal and Key Item support.
- Updated the way large quantities are displayed to be more compact and easier to read.
- Fixed inventory quantities not updating when not moving the full stack.
- Fixed Collectable items being stacked when they shouldn't.
- When listing items at your retainer's Market Board, items that can't be sold on the market are greyed out and locked.
- Fixed a crash when dismissing the last recently-looted item.
- Fixed a crash when closing the inventory window with the recently-looted strip still visible.
- Add option to display bag stats in the server info bar.
- Fix sublabel spacing in item/currency picker.
- Add changelog window to the configuration window.

### DIFF
--- a/testing/live/AetherBags/manifest.toml
+++ b/testing/live/AetherBags/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Zeffuro/AetherBags.git"
-commit = "6a0cca47e49c4945954e292e40078309a8f8e094"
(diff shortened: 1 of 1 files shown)


Zeffuro:
AI was used/asked to help trying to track down one obscure bug, the fix was written manually and has been verified fixing the issue.

nathanctech (approved):

PR #8662, opened 2026-05-15
ChatTwo

Infiziert90:
This is the first in a series of updates :)

### Popouts with input
- Popouts can now have input fields, with simplified channel switch or even tell targets

image

### Tell targets
- Make a tab fully focused on a single person

### HTML export
- Include the main svelte project 
- Export will be released with the next update

image

### Misc
- Fix a small memory leak
- Improve performance by not converting colors every frame
- Adjust button sizes
- Update locs


Infiziert90:
Note:

3800 lines of the diff are the standalone svelte project for creating the HTML export file

nathanctech (approved):

PR #8684, opened 2026-05-17
[Testing] Mini 2.1.1.7

Limiana:
- New option: Disable rendering while minimized
- - Disables rendering while game is minimized, which saves GPU resources


Limiana:
Reworked this process and made it so multiple plugins can use this function without fighting with each other

by the way if you don't want this function let me know

reiichi001:
Hi there! Apologies on the wait while this was discussed internally.

I wanted to get some more PAC input as a render bug could result in larger issues. After getting some additional eyes on the code, we (PAC/PAC-CodeReview) feel like it's probably safe, but would still like to see how testing goes, and if we get any feedback/support posts about the feature.

reiichi001 (approved):

PR #8685, opened 2026-05-17
TidyChat v2.3.1.0

Kagekazu:
- Fix DTR bar not updating when u change instance.
- Fix Battle currency being outdated and not hiding properly.
- Add option to hide the error from fate syncing.
- Remove glamour spam settings, since its in base game now.
- Fix `Another player has cast his/her loot` showing when it should not.
- Add better treasure dungeon messages


Kagekazu:
bleatbot, rebuild

PR #8701, opened 2026-05-18
HUDUnlimited

MidoriKami:
Add additional code safety.


MidoriKami:
My IDE has been progressively removing extra whitespace in files that I touch, for a cleaner diff I recommend looking at the Semantic view.

reiichi001 (approved):

PR #8715, opened 2026-05-21
Update plugin Clock 1.0.6.8

LatencyBryer:
### Update 1.0.6.8 - Focus: Timestamp feature

The native text-format hook is registered through Dalamud signature scanning and enabled during service construction.

The detour still:
- checks whether the requested addon text ID is the chat timestamp ID
- falls back to the original formatter when custom timestamps are disabled
- falls back to the original formatter for unrelated text formatting calls
- returns the custom timestamp string only when the plugin settings allow it

The custom timestamp output still respects the previous existing configuration options:

- 12-hour/24-hour display
- optional `AM/PM`
- configured timezone offset
- optional custom timestamp color;

> `Match Channel Color` option was temporarily removed since I couldnt get it to work properly/consistently.

The cached `Utf8String` pointer is still reused between calls but the code now reads more clearly around allocation, deletion and replacement.
Disposal behavior remains the same: the hook is disabled/disposed and the cached native string is deleted when the service is disposed.

**This update should solve any previous plugin incompatibility issues.**

> Tried to be more informative to avoid any possible confusion.

### DIFF
--- a/stable/Clock/manifest.toml
+++ b/stable/Clock/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/seventity7/Clock.git"
-commit = "4969541ea139f6b453171bfdae9aa2169c206693"
+commit = "aeb223c7ec5bb2a27c3cc56517346585ebb0164e"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
--- a/stable/Clock/manifest.toml
+++ b/stable/Clock/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/seventity7/Clock.git"
-commit = "96db6b84eeffc07bb1dd94485208cdcbcdbca19d"
+commit = "4969541ea139f6b453171bfdae9aa2169c206693"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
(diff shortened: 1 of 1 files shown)


nathanctech:
Compile error:

` [04:22:51 INF] [CONTAINER] /work/repo/Plugin.cs(123,30): error CS1061: 'ChatTimestampService' does not contain a definition for 'ApplyConfiguration' and no accessible extension method 'ApplyConfiguration' accepting a first argument of type 'ChatTimestampService' could be found (are you missing a using directive or an assembly reference?) [/work/repo/Clock.csproj]`

Please make sure your code compiles locally.

LatencyBryer:
Yep I'm fixing it and will rebuild in a moment.

LatencyBryer:
Updating commit since I did a **Preventive cleanup** and **hardening** Update.

### Hook active behavior
- The hook is no longer treated as always-on behavior.
  - Enabled only when **Show custom timestamp** is active and disabled again when the feature is turned off.
> This should reduce unnecessary runtime impact
```csharp
chatTimestampService.SetEnabled(Configuration.ShowCustomTimestampInChat);
```
- The service now updates its internal state when related settings change instead of doing extra checks every time a timestamp is formatted.

### Repeated logic out of the detour
The detour was cleaned up so it does not repeatedly sanitize config, resolve timezone state or rebuild unnecessary option data for every chat timestamp.

Settings such as:

- selected time format
- AM/PM suffix behavior
- timestamp timezone
- custom timestamp color

are prepared through the service update path and then consumed by the formatter.
This makes the hook smaller and keeps the risky native detour focused only on formatting the timestamp.

### Native timestamp target centralized
The native timestamp IDs are now centralized as named constants instead of being scattered through the logic:
```csharp
private const uint NativeTimestampTextId = 7840;
private const uint NativeTimestampTextIdWithSeconds = 7841;
```
These values are still required by this native formatter approach but keeping them named should make the intent clearer.

> **I am more than open for any sugestions or advice about the feature;**

nathanctech (approved):

PR #8718, opened 2026-05-21
TidyChat v3.0.0.0

Kagekazu:
Rewrote the whole plugin to use messageID instead of regex where possible.

- Bug fixes:
  - Chat history evicted newest instead of oldest (Stack → Queue/FIFO).
  - Crash fixes: invalid whitelist regex, single-word player names, null UIState derefs.
  - Thread safety: locking on log-message sets + chat history.
  - Dispose() now unsubscribes UI events, removes windows, releases the DTR entry.
  - UI: ChatHistoryTab saved every frame; Inverse Mode couldn't be enabled; bad EndTabBar.
  - Whitelist: Allow overrides emote hiding & beats Block; works in disabled channels; empty-name no longer matches all; added match modes.
- Perf: cached AllRules + whitelist/name regexes (were rebuilt per message).
- Cleanup: deleted dead GetDutyName.cs, merged 3 DTR settings, L10N format/fallback fixes, SmolMode off-by-one, rule typo/copy-paste fixes.
- API: SetPlayerName uses ObjectTable.LocalPlayer (deprecated IClientState.LocalPlayer).
- New option : server announcement filter

### DIFF
--- a/stable/SimpleTriggers/manifest.toml
+++ b/stable/SimpleTriggers/manifest.toml
@@ -1,10 +1,13 @@
 [plugin]
 repository = "https://github.com/Brolijah/SimpleTriggers.git"
-commit = "6276375678edae7208f4717401d3865be8fdad16"
+commit = "31575ef24a5b0d11d394b65aed223497cd8c7c8e"
 owners = ["Brolijah"]
 project_path = "SimpleTriggers"
 changelog = """
-## SimpleTriggers v1.0.2.3
-* Minor improvements.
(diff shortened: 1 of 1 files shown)


nathanctech:
bleatbot, rebuild

nathanctech:
Something is wrong here...the force push erased your change. @Kagekazu

Kagekazu:
> Something is wrong here...the force push erased your change. @Kagekazu 

Oh, I'll resubmit

PR #8719, opened 2026-05-21
Update SimpleTriggers v1.0.4 (stable)

Brolijah:
* Replaced DECtalk with DECtalkMini
* Fixed scaling and alignment in various places
* Added a new option for trigger responses: popup toasts/gimmicks


Brolijah:
Quick fix before merge: 
* Fixed non-fatal UI error when removing a category
* Trims new-lines from messages before checking against trigger expressions

nathanctech (approved):

PR #8722, opened 2026-05-22
ChillFrames

MidoriKami:
UI Touchup + Code Cleanup


MidoriKami:
Big diff is from whitespace removal, recommend using Semantic view

nathanctech (approved):

PR #8750, opened 2026-05-25
SortaKinda

MidoriKami:
Reworked plugin, now with greater capabilities, and significantly better performance.

Not compatible with previous versions of the plugin. All presets prior to v3 are not compatible.

Several new filtering options, new ordering options, ordering can now be done as a tiered list of tiebreakers.

New Filters or Ordering options can be easily added over time.

Replaces the games "Sort" context menu option to trigger SortaKinda sort if any rules are defined for the target inventory.

Immediately updates UI upon completing a sort.

### DIFF
--- a/stable/SortaKinda/manifest.toml
+++ b/stable/SortaKinda/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/SortaKinda.git"
-commit = "e48186585a074db0b0fc216949f7e7d713ba995d"
-owners = ["MidoriKami"]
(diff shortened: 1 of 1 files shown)


MidoriKami:
Diff is actually small.

Actual diff from Testing : https://github.com/MidoriKami/SortaKinda/compare/22e5afff783df8a015a77b0668a3aae61a2c3cd6..1efd44573afd8507fd8c0f96dd2869dbf5337abc

This is a channel move + 2 added orderings

nathanctech (approved):

PR #8755, opened 2026-05-26
XIVInstantMessenger 1.5.0.20

Limiana:
- Fixed an issue where messages in forays could not be sent
- Vertical tabs will now be scrollable independently of the chat

### DIFF
--- a/stable/Messenger/manifest.toml
+++ b/stable/Messenger/manifest.toml
@@ -5,7 +5,7 @@
 # must be accessible (i.e. no `git` or `ssh` URLs, as the agent cannot clone those)
 repository = "https://github.com/NightmareXIV/XIVInstantMessenger.git"
 # The commit to check out and build from the repository.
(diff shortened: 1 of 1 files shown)


nathanctech (changes requested):
I must have missed this for the API 15 push and I apologize. Please remove this from your csproj:

`<PackageReference Include="DalamudPackager" Version="15.0.0" />`

This could break in the future.

Limiana:
> I must have missed this for the API 15 push and I apologize. Please remove this from your csproj:
> 
> `<PackageReference Include="DalamudPackager" Version="15.0.0" />`
> 
> This could break in the future.

I think that'll only work when using Dalamud custom SDK

Do you plan to move away from packager / do you want plugin to be switched to use Dalamud SDK?

nathanctech:
Unless you can't for some reason, definitely use the SDK as it handles the packager and that sort of thing for you.

Limiana:
It is done now

nathanctech (approved):

PR #8764, opened 2026-05-27
Fatal Frame 1.3.0.1

Glyceri:
Fixes achievements not capturing properly.


Haselnussbomber:
Curious, why not use IUnlockState for this? It hooks the same function.
I guess I won't get an answer here, so since I don't actually care, I'll mark this as outdated.

perchbirdd (approved):

Glyceri:
> Curious, why not use IUnlockState for this? It hooks the same function. I guess I won't get an answer here, so since I don't actually care, I'll mark this as outdated.

I don't think the unlock state is fully up to date yet with what I track.

PR #8771, opened 2026-05-28
[Testing] AkuTrack v0.0.33

amki:
Mainly updates AkuTrack to work with Dalamud API 15.

Configuration is now actually loaded after saving and picking a color works, the map can be centered on a player and there is an optional camera cone.


nathanctech (changes requested):
Please remove the DalamudPackager and ReproducibleBuild packages from your csproj as the SDK handles these.

amki:
I repaired that, thank you.

nathanctech (approved):

PR #8796, opened 2026-05-31
Update manifest for CurrencySpender

Blackcatz1911:
# Changelog

### Added
- Fashion accessories can now be considered a collectable.
- Settings to hide the UI in various scenarios.
- Custom repository plugin support for various locations.
### Fixed
- Spelling errors.
### Changed
- Updated for Dalamud API v15.
- Updated ECommons.
### Removed
- Kofi Banner.

### DIFF
--- a/stable/CurrencySpender/manifest.toml
+++ b/stable/CurrencySpender/manifest.toml
@@ -2,13 +2,12 @@
 repository = "https://github.com/Blackcatz1911/CurrencySpender.git"
 owners = ["Blackcatz1911"]
 project_path = "CurrencySpender"
-commit = "f08fd02320adce9ca1b1749f34823935976a1c71"
+commit = "69612732fa9cbb139d5fa67de3aff39f6dfa9ed3"
 changelog = """
 ## 1.2.6
 ### Added
 - Fashion accessories can now be considered a collectable.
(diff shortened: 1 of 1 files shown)


nathanctech (changes requested):
Hi, please do not reflect into other plugins unless absolutely necessary. If using IPC, you can check if it's available or not without checking for the existence of the plugin via [HasAction/HasFunction](https://github.com/goatcorp/Dalamud/blob/dcbc5bf92abef8cf98309655f23bea01a38cb085/Dalamud/Plugin/Ipc/ICallGateSubscriber.cs#L10-L17) on the subscriber or attempting to use the IPC and failing gracefully if it doesn't exist.

Blackcatz1911:
Removed the Reflector class and just check if the plugin is enabled. That should suffice for my use case

nathanctech:
> Removed the Reflector class and just check if the plugin is enabled. That should suffice for my use case

Thanks - one more quick thing, please remove named references to custom repo plugins in your changelog as we don't allow those to be mentioned. You can mention plugin support, but it can't be explicit if it's a custom repo.

nathanctech (approved):
Thanks. I edited your description too.

PR #8816, opened 2026-06-03
Update plugin Clock 1.0.7.0

LatencyBryer:
### Changes
- Added recurring alarm handling with repeat reset after alarm creation
- Added optional main overlay text for:
  - Next scheduled alarm
  - Current-day detected maintenance
- Added configurable text size, vertical positioning, text color, shadow color, and shadow offset for the new overlay reminder texts
- Added `/clock <timezone1> to <timezone2>` timezone comparison command
- Added optional `/clock` command suggestion popup with filtered command results while typing
- Added `Command suggestion` behavior setting under `General`
- Added appearance shortcut icons that jump to relevant appearance sections
- Updated Appearance presets to preserve user layout, size, position and format while applying visual styling only
- Removed the old Profiles preset section in favor of the existing Appearance preset flow
- Updated alarm validation to prevent creating alarms in the past
- Updated alarm history **date display** to use a shorter **English-style format** _(I was displaying it as Brazilian format before)_
- Added localization entries for new settings, labels, tooltips and command descriptions

### Fixes
- Fixed timezone conversion using the wrong source timezone label in some cases
- Fixed some untranslated tooltips/texts

### DIFF
--- a/stable/Clock/manifest.toml
+++ b/stable/Clock/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/seventity7/Clock.git"
-commit = "5b3499691aa5dcdb6a48f7e200131bccc1605b7e"
+commit = "6e2838f384a6a007e27566bc875be292ea60dcb4"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
--- a/stable/Clock/manifest.toml
+++ b/stable/Clock/manifest.toml
@@ -1,27 +1,25 @@
 [plugin]
 repository = "https://github.com/seventity7/Clock.git"
-commit = "ddd614a73b75fd9278ac83abd6bc94a47d254a35"
+commit = "5b3499691aa5dcdb6a48f7e200131bccc1605b7e"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
 changelog = """
-This update improves how Clock detects and announces upcoming FFXIV maintenance.
+## What's new
+- Alarms can now repeat instead of being one-time only
+- Added overlay reminders
+  - Clock can now show the next alarm on the main overlay
+  - Clock can now show today's detected maintenance on the main overlay
+- When typing `/clock`, a small suggestion list can now appear with available Clock commands
+  - This can be enabled or disabled from the `General` tab
+- You can now compare two timezones using commands like:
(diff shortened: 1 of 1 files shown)


nathanctech (changes requested):
Hi,

Your `GetActiveChatTextInput()` is very prone to breaking and we heavily discourage use of signatures that exist in CS for you to use.

Another developer provided some example code you can use to detect if you're within the chat input component and accomplish what this function does:

```
        // RaptureAtkModule.Instance() is guaranteed to be valid after GameInit
        var raptureAtkModule = RaptureAtkModule.Instance();
        ref var textInput = ref raptureAtkModule->TextInput;
        if (textInput.TargetTextInputEventInterface is not null && raptureAtkModule->IsTextInputActive()) {
            // There is an active text input

            // AtkStage.Instance() is guaranteed to be valid after GameInit
            var inputManager = AtkStage.Instance()->AtkInputManager;
            if (inputManager is not null) {
                var focusNode = inputManager->FocusedNode;
                if (focusNode->GetNodeType() is NodeType.Text) {
                    // Chat is active and this node's parent node is a AtkComponentNode with AtkComponentTextInput Component
                }
            }
        }
```

Please ask if you have any questions about how something should be implemented as some methods may be worse than others, especially unreviewed code that exists on custom repositories.

LatencyBryer:
Hi! About the feedback regarding `GetActiveChatTextInput()`:

- Reworked chat input detection to avoid the previous signature/vtable-based approach
- Removed the dependency on `ISigScanner` for command suggestion input detection
- Updated `Plugin.cs` so `GetActiveChatTextInput()` now uses the suggested ClientStructs path:
  - `RaptureAtkModule.Instance()`
  - `RaptureAtkModule->TextInput`
  - `RaptureAtkModule->IsTextInputActive()`
  - `AtkStage.Instance()->AtkInputManager`
  - `AtkInputManager->FocusedNode`
  - `NodeType.Text`
  - Parent node validation against `AtkComponentTextInput`

The command suggestion popup now checks the active chat text input through exposed ClientStructs. This should make the implementation less prone to breaking and more aligned with the recommended. Thank you for pointing it out!

nathanctech (approved):
Thanks for making the changes. Note that a command hinter logic is pretty new territory as far as plugins so it will be important to monitor for reports of things not behaving properly, especially if users have other plugins that modify chat behavior.

PR #8817, opened 2026-06-03
FATEhelper v1.1

teechep:
added distance view and sort option (thanks to jade-doggerel on github)


nathanctech (approved):
FYI, you may want to use branches on your fork to avoid the odd git history in PRs in the future. No big deal, though!

PR #8825, opened 2026-06-04
QuickSymbols Compatibility Update

LatencyBryer:
**Focus: Plugin `Chat2` Compatibility**
- Added support for opening the Quick Symbols popup while using Chat2's ImGui chat input
- Added Chat2-aware keybind handling so the symbol popup can be opened through the Quick Symbols keybind when Chat2 is active
- Implemented symbol insertion into Chat2's current input state
- Updated Chat2 insertion cursor handling to use UTF-8 byte offsets
- The button `♥` is hidden while Chat2 is active to avoid duplicated windows/buttons

### Configuration UI
- Added additional help/tooltips near the keybind configuration controls
- Added Chat2-specific keybind guidance with separate recommended keybind lists for Windows/Linux

### Keybind and Input Handling
- Chat2 compatibility is separated from the vanilla chat path to avoid altering existing native chat insertion behavior

**Note:** This is not a colab with `Chat2` plugin and they developers have no responsability about this. I'm completely responsable for the compatibility work and I'm more than open to any suggestions/Feedbacks about it.

### DIFF
--- a/testing/live/QuickSymbols/manifest.toml
+++ b/testing/live/QuickSymbols/manifest.toml
@@ -1,10 +1,18 @@
 [plugin]
 repository = "https://github.com/seventity7/QuickSymbols.git"
-commit = "7179a5f0fe04d1c89d210cf20d3ab41f68e34f13"
+commit = "685ed6a0c962891c77de37889e6d1828f849fc5b"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
 changelog = """
-### Icon update
-- Changed plugin icon
+### What's New
+- Added support for **Chat2** plugin through the Quick Symbols keybind
+- Added a special popup behavior for better compatibility while Chat2 is being used
+- The button `♥` is hidden if Chat2 is active to avoid duplicated buttons
+- Added a new help tooltip next to the keybind setting with recommended keybinds for Chat2 compatibility
+  - Added separate recommended keybind lists for **Windows** and **Linux** users
+
+### Notes
+This is not a colab with `Chat2` plugin and they developers have nothing to do with this. Any feedback/suggestion/issues must be sent through `QuickSymbols` plugin on Dalamud.
+Thanks @infi for support directions/reply.
 """


nathanctech (changes requested):
Do not use reflection against other plugins - this is what IPC as for. ChatTwo exposes the information you need.

LatencyBryer:
Hi! I'm not gonna update the PR for now. I checked the current ChatTwo IPC and `TypingIpc` exposes the input state but I do not see a public IPC for inserting/appending text into ChatTwo’s input or updating its cursor/selection.
My plugin needs to insert symbols into the active ChatTwo input and keep the cursor correctly. So I need to know which supported IPC should be used for text insertion.

 Without reflection, the only available paths are vanilla/Atk or simulated input and both have caret/focus issues. I tested the vanilla/Atk path ChatTwo listens to. It updates the text, but it goes through ChatTwo’s `AddIfNotPresent` flow, which prevents repeated identical symbols and does not fully keep the visible InputText caret in sync.

 With the vanilla chat it works just fine but on the chat2 the "caret" of the textinput gets stuck on the initial insertion position even tho it insert the symbols in the correct order when pressed. Wich consequently causes wrong order of typing, making the user having to click on the end of the text to continue typing everytime after using a symbol, wich is not a deal.

 I have been making non-stop attempts to make it work properly but I didnt got any good result yet. I also tried seeking guidance on the Dalamud discord but havent received any proper reply yet for this problem. 

 I'll continue trying some things till tomorrow or so and will either close the PR, not releasing the update or if I have a good result I will update it for release.
This compatibility update was mainly been made because of user feedback/request, so if is something that can't be done, then sadly so be it.

 I'm attaching a file where I explain part of my code that is having the main problem. If any reviewer wants to take a look and maybe help me with some guidance would be much appreciated! If you can't is totaly fine as well.

[QuickSymbols insertioncaret.md](https://github.com/user-attachments/files/28616730/QuickSymbols.insertioncaret.md)

> note: Before the PR I made some questions to `ChatTwo` developer wich he was able to answer some of them and some not. Since this is not a colab and just a update by user feedback, I dont want to keep going after the developer frequently asking for stuff.

nathanctech:
I'll close this for now until you have code ready. As I mentioned in another PR, if you wanted to experiment with certain features you can use the "test" branch. I don't know enough about chat input to assist with this feature, sadly.

PR #8832, opened 2026-06-05
Update plugin Clock 1.0.7.1

LatencyBryer:
Added an opt-in **Chat Time Conversion** feature that detects timezone-based mentions in chat and shows the converted result on hover.
The feature is marked as experimental because it modifies the displayed chat `SeString` and may interact with other chat-related plugins
image

### Main Changes
- Detects common time formats with explicit timezones, including single times and ranges such as `8PM BRT`, `7PM-10PM EST`, `7PM @ 10PM EST`, and `07 PM - 10 PM GMT`
image 
image

- Supports AM/PM text and FFXIV AM/PM symbols
- Shows a compact hover tooltip with the converted time
- Adds a `Create Alarm` action when available
image

- Uses the first time in a range as the alarm target, while keeping the full range visible in the tooltip
- Reads date context from the same chat message for alarm creation, such as `June 12th`, `12th June`, or localized month names where supported

### Settings
- `Enable chat time hover`: enables chat time detection and hover conversion
- `Show alarm setup option`: controls whether the `Create Alarm` action can appear for valid times
- `Target timezone selector`: chooses the timezone used for displayed conversions
- `Tooltip duration`: controls how long the tooltip remains visible
- `Test button`: sends local test messages that exercise single-time parsing, range parsing, date context, hover display and alarm creation. _(Testing pourposes)_

### Safety/Compatibility Notes
- The feature is disabled by default and requires a explicit opt-in confirmation
- A one-time experimental warning is also shown before first activation and is only accepted when the user clicks `Yes`
image

- Existing chat messages **are not** consumed, resent or modified at the source; only the displayed `SeString` is adjusted to add the hover link
- The hover popup is drawn as a normal Dalamud window and is kept separated from the game chat input

This is a feature that fits the plugin main idea pretty well. This is was a quite tricky feature to create and it's not perfect. It is maked as  Experimental with multiple warnings on pourpose to avoid any "unaware" issues.

I'm more than open for feedbacks and suggestions, I'm sure there must be some things that could be done better and I'm more than open to listen.

### DIFF
--- a/stable/Clock/manifest.toml
+++ b/stable/Clock/manifest.toml
@@ -1,25 +1,27 @@
 [plugin]
 repository = "https://github.com/seventity7/Clock.git"
-commit = "6e2838f384a6a007e27566bc875be292ea60dcb4"
+commit = "6c7da0981d18cb9f7cb89f83ea4e99c5280c8531"
 owners = ["seventity7"]
 maintainers = ["seventity7"]
 project_path = "."
 changelog = """
-## What's new
-- Alarms can now repeat instead of being one-time only
-- Added overlay reminders
-  - Clock can now show the next alarm on the main overlay
-  - Clock can now show today's detected maintenance on the main overlay
-- When typing `/clock`, a small suggestion list can now appear with available Clock commands
-  - This can be enabled or disabled from the `General` tab
-- You can now compare two timezones using commands like:
-    - `/clock EST to BRT`
-- Appearance shortcuts were added to make related settings easier to find
+## What's New
+Added an experimental **Chat Time Conversion** feature for the in-game chat messages
+When enabled, the plugin can detect date/times in chat, including single times and time ranges, ex: `8PM EST``7PM-10PM EST`, then show the converted time when hovering over it
 
-### Improvements
-- Alarm creation now prevents setting alarms for dates or times that already passed
-- Presets now focus on visual style without changing the user's current layout, size or position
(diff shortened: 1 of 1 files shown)


LatencyBryer:
Just a quick observation. I left a "Warning" log on the code from when I was debbuging the feature. is a diagnostic message for the Chat Time Conversion feature; it records whenever a user clicks a detected time in chat and the conversion popup is opened, including the original detected text, the converted result, the target timezone and the internal local/UTC time values used to decide whether actions like Create Alarm should be available.
So if you have any issue while Testing the feature, please send me the log from `/xllog`, the one starting with `[Clock.ChatAlarmSetup]`. 

This will be removed or changed from WRN before merge if approved.

nathanctech:
I recommend utilizing the "test" branch for experimental code so testers can opt-into changes that might not be fully stable. You can maintain both a testing and stable branch of your plugin if you want to. If you want this in stable, I can merge as-is.

LatencyBryer:
I totally forgot about the Test branch but I still want it to be merged at the Stable version. 
I think I have put enough Warnings disclaiming the `Experimental` part, so it should be fine. And the feature is already coded in a way to make the user full aware of it and have the choice to use it or not after acknowledging the possible instability/issues.
> edit: Also, I dont think there is any possibility of big issues like Crashing the game or something like that.

nathanctech (approved):

PR #8834, opened 2026-06-06
[Retainer Alerts] Updated for better usability.

Ibcrootbeer:
Updated to make it so the alert window will not show unless the player is on their home world.


nathanctech (approved):
Thanks. I advise creating a new branch on your D17 fork to avoid git shenanigans like you see here. :)

PR #8837, opened 2026-06-06
Updated ARRP to 0.1.5.0

NetStyx:
# Functional changes
- added the ability for actors to head track the players position
- added the ability for actors to play existing timelines directly
- added the ability for actors to sit or sleep on appropriate furniture when they are close enough
  - remember to `stand up` after you `sit`.
- updated the debug ui to not depend on the main window being open
- updated the different pickers to provide more data
# Technical changes
- removed the custom speech bubble code and switched over to existing implementation
- removed custom emote code and switched over to existing implementation
- fixed some issues with character file parsing


nathanctech (approved):
This is probably fine, please keep an eye on behavior.

PR #8856, opened 2026-06-09
Ascian Music Player 1.2.1.6 (Testing)

GayPotatoEmma:
## Fixes
- Add fallback to DirectSound if no WaveOut device can be initialized (hopefully)


GayPotatoEmma:
Large diff due to testing being updated from old version. Here is the actual diff from the current stable version to this
https://github.com/GayPotatoEmma/Ascian-Music-Player/commit/961daee596a160187eb2d35b6e40185690259060

nebel (approved):

PR #8868, opened 2026-06-11
Blue Mage Helper 3.4.1.0

harbingerftw:
- Fix showing head markers on mobs when already learned.
- Added the ability to show head markers all the time


harbingerftw:
Actually I just noticed I left in the debug open window lines, let me remove those first (well be several hours)

nathanctech (approved):

PR #8876, opened 2026-06-12
Update plugin QuickSymbols

LatencyBryer:
Last icon image I sent using 512x512 was showing up full of artifacts in the Installer window so I'm updating the Icon image again since I found out that is better to use 128x128 and below instead of 256x256 and above because of how the downscaler works.

Also updated the Links & etc to my new github Username.

No code changes.


nathanctech:
I think your URL is incorrect there.

nathanctech:
Thanks, I don't see a new icon listed in your PR and you've only added this to testing. Just wanted to make sure that was your intent!

LatencyBryer:
Oh, thanks. I changed in the repo with the commit and forgot to add it to the actual PR ç.ç I will be doing now, after that is just it.

edit: all good. First attempt didnt replaced for some reason, now is correct.

nathanctech (approved):

PR #8912, opened 2026-06-19
New Plogon Gaolbreak

sourpuh:
I've tested this about as thoroughly as I can, so sending it to the main repo at last. The alpha release in my custom repo has had 69+ testers and I've fixed all major known issues. See the [README](https://github.com/sourpuh/ffxiv_gaolbreak/blob/main/README.md) for plugin details.

The only parts of this plugin that I think may be controversial are:
1. The use of low-level ImGui functions to change draw ordering of various plugin windows which is necessary to layer with the Native UI.
2. DepthManager enables DepthBasedPriority on many Addons to split them into foreground and background capture textures. This is (mostly*) not visible to the user, but it has the potential to impact other plugins (no known issues).

\*  The full size BLM job gauge has a clipping issue with its enochian meter that I don't know how to fix yet.

**Disclosure:** Assist: Human-led. AI was used on demand for specific tasks.


sourpuh:
Just pushed all updates since I opened this PR.

* Update sigs for hotfix 2
* Replace depth prio system with new injected clipmask sentinels
  * Stop modifying depth priority like I mentioned in the PR description
  * Fix BLM enochian gauge clipping I mentioned in the PR description
  * Move ChatLog and NaviMap addons to BG
* Add double buffering to fix flickering with low FPS
* Add automatic killswitch on hook failure so game updates (hopefully) won't break UI rendering anymore.
* Enable in GPose
* Fix ReShade REST
* Fix text input focus when changing ImGui <-> Atk
* Enable in title screen & chara select

nathanctech:
Thanks for your contribution! This has been approved and will be added. Due to the low-level nature of this plugin definitely keep an eye on the testing and support channels in case people run into conflicts.

nathanctech (approved):

sourpuh:
Thanks, Nathan! :)

PR #8927, opened 2026-06-22
v1.0.10 updating to use new SDK version

ArcaneAj:
AI Usage: Hint (or whatever you call the default suggestions that vs community 2026 has)

Using the new SDK version

Fixed an issue where the manually set resources get wiped when you open a retainer/saddlebag even if you've opened it before.


ArcaneAj:
This has been tested locally, that's how I knew I had to update the new SDK version >.>

nebel (approved):

PR #8929, opened 2026-06-22
Update plugin QuickSymbols

LatencyBryer:
I’ve been receiving various requests for the `QuickSymbols` plugin to work with other plugins, specifically the `Chat2` plugin.
This feedback is coming in not only through the feedback channel on the Dalamud server but also via my DM's in Discord.

I don't mind receiving those feedbacks but since the request is something that they shouldn't be asking to me and instead to `Chat2` developers, I decided to put a disclaimer explaining it so people don't think I'm ignoring them.

I’ve added an icon with a tooltip to the `QuickSymbols` settings to provide information about plugin compatibility.
image
> Just note that even tho these type of request shouldn't been sent to me and instead Chat2 devs, I still never said to the user's to go bother them. I just wanted to point out a explaination for they feedback.

### DIFF
--- a/stable/QuickSymbols/manifest.toml
+++ b/stable/QuickSymbols/manifest.toml
@@ -1,18 +1,9 @@
 [plugin]
 repository = "https://github.com/LatencyBryer/QuickSymbols.git"
-commit = "d5f35580462cd7cc0979c72d5d478f88da53014a"
+commit = "c53f089146f9e42b189bfe6df81702e40400e814"
 owners = ["LatencyBryer"]
 maintainers = ["LatencyBryer"]
 project_path = "."
 changelog = """
-### What's New?
-- Fixed "♥" button alignment/centering
-- The "♥" button now appears in more native windows:
-  - User Macros
-  - Strategy Board
-- New **optional** feature: Automatic Text-To-Symbol
-  - Automaticaly convert known texts to symbols, like <3 -> ♥
-  - Create your own Text-To-Symbol in the config
-Note:
-"Text-To-Symbol" feature is intended for vanilla chat only and will work ONLY with it.
+- Added a "Compatibility Disclaimer" on config window
 """


nathanctech:
Wouldn't it be better to give "copy to clipboard" support for now until IPC support is added for Chat2? I'm sure the users would be happy to have something in the meantime.

LatencyBryer:
> Wouldn't it be better to give "copy to clipboard" support for now until IPC support is added for Chat2? I'm sure the users would be happy to have something in the meantime.

I considered clipboard support but I don't think it would be good to have instead of not having anything.

If it only copies to clipboard, the user still has to manually paste every symbol, which not only becomes awkward when using multiple symbols or when the picker is opened through a keybind during typing but it also defeats the proposal of the plugin in the first place.

It also could possibly create multiple side effects as well, like:

* It would overwrite whatever the user currently has in their clipboard
* It changes the behavior depending on whether the target is vanilla chat or Chat2
* It could feel unreliable because the plugin would say a symbol was selected but nothing would appear unless the user manually pastes it _(And yes, is not viable to show a feedback text for when the user Copy it to they clipboard.)_
* For people using QuickSymbols frequently, copy/paste is a much slower way than direct insertion

So as a temporary fallback it is possible but I worry it would feel more like a poor workaround. I would rather keep the current direct insertion behavior for vanilla chat and a disclaimer about other plugins compatibility instead of adding clipboard behavior that may confuse users or cause unwanted side effects.

nathanctech:
I would certainly say it's viable to allow clipboard support because a lot of people already use macros to store all the glyphs and manually copy-paste from there to where it's needed. I would highly consider this option to be opt-in so users of non-native chat can use your plugin. You can also provide feedback by maybe say changing the icon they click to a checkmark or something. I think you're worrying too much about usability issues from a feature that a user would have to toggle themselves.

Either way, I will merge this but I will highly encourage giving additional support otherwise users will still bug you until it's implemented.

nathanctech (approved):

PR #8951, opened 2026-06-28
[Testing] DailyDuty

MidoriKami:
Fix Search Not working (forgot to re-implement it tehe)

Fix todo panel not hiding completed entries.


MidoriKami:
Is actually a small diff, this is after crowdin added localization files for all the other langauges (currently in english)

nathanctech (approved):

PR #8960, opened 2026-06-29
HUDUnlimited

MidoriKami:
When HUDU Can't find a node, it will now only log the warning once per session.

Some nodes don't exist at all times, and so seeing a warning for some nodes might be normal.


MidoriKami:
Recommend semantic view of the diff, IDE removed some empty whitespace characters.

nathanctech (approved):

PR #8975, opened 2026-07-04
HUD Manager 2.5.27.0

nebel:
HUD Manager gives you total control over your HUD layouts:
- Save infinite layouts
- Edit your HUD with precision
- Automatically swap between layouts based on conditions
- Make child layouts that only affect desired HUD elements
- Change visibility of job gauges
- Export and import layouts from text

After 6 long years, HUD Manager is no longer testing-exclusive. Kept you waiting, huh?


nebel:
Note: This is the same as the current `testing` release.

nathanctech (approved):

PR #9013, opened 2026-07-12
NoTankYou

MidoriKami:
Various system upgrades, performance improvements, and stability improvements.

Reworked config ui a little, now more sleek and modern.


MidoriKami:
This is a release from testing -> stable, with a KTK stability update

(This ktk version is the same as the DailyDuty and VanillaPlus updates that I also pushed with this)

nathanctech (approved):

PR #9035, opened 2026-07-19
Add PopotoVox

michaelcurry:
### PopotoVox — local NPC voice acting

PopotoVox gives FFXIV NPCs distinct synthetic voices, rendered entirely on the player's machine: dialogue capture → per-NPC voice casting → local TTS → mixed playback. No cloud services, no accounts, no telemetry.

**Voice Engines** (player-selectable): Piper and Kokoro run on CPU; the optional "Ultra" tier (VoxCPM2) designs a unique voice per NPC and needs an NVIDIA GPU.

**Runtime downloads & helper processes** (flagging this proactively): voice models and engine runtimes are large, so they are not bundled — the plugin downloads them on demand and runs the engines out-of-process. How that's kept safe (all in `plugin/Security/`):

- The plugin can only ever download files from a fixed list that is compiled into the plugin DLL and signed (ECDSA P-256). If the signature check fails, downloading is disabled entirely — there is no "download anyway" path.
- Every file on that list is pinned to an exact SHA-256 hash and byte size, and is verified before it is installed.
- Downloads are HTTPS-only and limited to an allowlist of hosts (github.com, githubusercontent.com, huggingface.co and its CDNs). Redirects are followed manually so every hop is re-checked against the same allowlist.
- The engines themselves run as separate child processes, so the plugin assembly contains no native code. They communicate over stdin/stdout, except llama.cpp (used for voice casting), which listens on `127.0.0.1` only. All helpers are killed when the plugin unloads, and the Python host also watches the game's process and exits if it dies.
- The plugin installs no game hooks — dialogue capture is `IAddonLifecycle` events plus object-table reads.

Disk use is shown before download (CPU tiers ≈ 0.5–1.5 GB; Ultra ≈ 12.6 GB, so it is opt-in with an explicit hardware check). Uninstalling removes everything.

**AI usage disclosure (per the AI Usage Policy): Copilot.** The code was written with an AI assistant (Claude); I planned, directed, and reviewed the work and validate every change in-game. I stand behind the implementation and am happy to walk through any part of it. The plugin icon and gallery images are AI-generated (disclosed in the plugin description).

License: GPL-3.0 (the CPU voice engines include espeak-ng, which is GPL). Full third-party ledger in `docs/LICENSES.md` in the repo.


nathanctech:
bleatbot, approve

nathanctech:
You appear to have the wrong owner set. Either set the owner to you (the PR submitter) or you'll need to re-submit using the account you wish to be the owner of the plugin. You can also add your name so both are in the owners list.

michaelcurry:
hummm  🤔  build error of `Not your plugin` I have a feeling this is because this repo is owned by my Org `kernelcurry` yet I am the one who is opening the PR for the plugin?  OR is there something else I am missing here?

michaelcurry:
Looks like the build got further 🥳, but then stopped as there are new dependencies that have not been reviewed before?

If there is anything I can do to help speed this along, please let me know.

nathanctech:
It looks like your plugin is attempting to use the WindowsDesktop SDK which isn't supported by our build system (at the moment). If you have a specific need for this, please let me know. NAudio tends to include Windows-specifc APIs when you just add everything which could be the cause.

michaelcurry:
`NAudio` is now `NAudio.Core` + `NAudio.WinMM` only, which removes the `WindowsDesktop` reference. Also dropped two runtime-pack pins the helper never actually needed (`Microsoft.WindowsDesktop.App.Runtime.win-x64`, `Microsoft.AspNetCore.App.Runtime.win-x64`).

**TL;DR on the remaining new/unreviewed packages:**
- `org.k2fsa.sherpa.onnx` ×9 — the engine for the Kokoro voice tier. Only win-x64 is used and shipped, but upstream's managed package hard-depends on all 8 platform runtime packages and publishes no win-x64-only split, so they all land in the lockfile.
- `SharpZipLib` — extracts the Kokoro model archive; upstream distributes it as .tar.bz2 and the .NET BCL has no bzip2 support.
- `Microsoft.NETCore.App.Runtime.win-x64` + `Microsoft.NETCore.App.Host.win-x64` 10.0.10 — the .NET runtime itself, straight from Microsoft. The TTS helper runs as its own process outside the game (so an engine crash can't touch FFXIV), which means it can't use the runtime Dalamud injects — it bundles its own copy instead. These two packages are that copy, and they're listed in the lockfile because the offline build server only downloads packages the lockfiles name.

michaelcurry:
I was able to remove the nine `org.k2fsa.sherpa.onnx` packages entirely: the TTS helper now calls sherpa-onnx's C API directly, and its two native DLLs install the same way as Piper — downloaded from the upstream k2-fsa release, SHA-256-pinned in the plugin's signed manifest (byte-identical to the DLLs the NuGet package shipped). the remaining new packages are just SharpZipLib and the two Microsoft .NET runtime packs.

nathanctech:
Quick thing I wanted to bring up: Is there anything this plugin does that https://github.com/karashiiro/TextToTalk doesn't? I only ask because there was another plugin recently that the dev added support for so they didn't need to maintain their own plugin.

michaelcurry:
I actually used this plugin for a while, had it setup with a cloud api solution , and it worked for basic tts, but very much seemed like you needed to be pretty technical to set it up the way you wanted.

I went to the drawing board with a few goals:
1) Setup experience should be as automated as possible, and settings should be curated, giving the best experience to general users.
2) everything should be local. I wanted to explore what could be done with local models. It has been a fun ride so far. 
2.5) I wanted to make a “voice casting” flow that would allow for NPCs to have distinct voices and accents, giving a bit more depth/life to the world.

michaelcurry:
Updated to v0.9.14 (08e8931). Some UI updates: the NPC list now shows one entry per character instead of duplicates (the game gives the same NPC a new ID in each zone/scene), and NPCs with no name get a readable description instead of a blank row.

No dependency changes.

nathanctech:
Doing some code review and testing...notes:

[Build] Plogon doesn't seem to like something still. NAudio appears to be outdated, try bumping to 2.3.0 to see if that fixes it. Otherwise, you might consider trying the 3.0 prerelease.

[Blocking] The plugin refuses to download manifests. I built it locally since it's not currently working with Plogon: "Asset manifest signature verification FAILED. Refusing to use it (PRD §8.1)." I've tried rebuilding a couple of times with no success. I don't think you need this as the underlying manifest cannot be altered after being built and all plugins are distributed via Dalamud's infrastructure.

Side note, why do you have license checks in here? The plugin cannot be altered by an end-user so the checks seem pointless, in my opinion.

After bypassing checks...continued.

[Concern] The LLM appears to use significant CPU even on my test machine (7900x3d, AMD GPU). While parsing, game stutters to about 30-40 FPS on a lightly loaded zone. This is with the High preset recommended by the plugin.

[Concern] There appears to be no way to stop a voice from playing even after the text is advanced or dismissed.

[Concern] The code and UI feel entirely AI-generated complete with the telltale dashes everywhere. Have you gone through and proofed all the code?

[Concern] The plugin appears to assume English-only and is not at all localized. If this is intentional, you'll want to make note of this as we have a large population of users who don't play in English. I was unable to test what happens if the LLM is fed a different language's text, but I assume it will not function correctly. Also, the plugin appears to be built for Windows only and this should also be called out as we officially support Linux.

Overall, this is a pretty large codebase and difficult to navigate...so I may have missed something. I have concerns over how difficult it will be to maintain this project as time goes on particularly when underlying APIs (like ollama) change since everything appears to be written by AI. But, we'll start with the concerns I have above and go from there.

michaelcurry:
Hey man, thanks for taking the time to read Through the code and give feedback / debugging notes ^_^  All valid points, I am continuously working on this, and if we want to hold off on publishing until  some of these concerns are resolved, all good. (e.g. Japanese language voice generation support, Better testing with AMD hardware, and Linux support)  At least that way, a majority of users would be able to play out of the gate.  

It might take some time for me to resolve all of these items...  Would you like to close the PR for the time being?

nathanctech:
Yes, I feel like it needs some work before it's ready, particularly performance as people will definitely complain about lag when it does its thing. In its current state, it doesn't work correctly with the manifest verification issues anyway.

PR #9065, opened 2026-07-27
Add Regions of XIV to testing/live

Yesanith:
Adding Regions of XIV, a new plugin, to testing/live.

FFXIV announces zone changes on screen but changes your sub-area silently the
only feedback is the small text above the minimap. This surfaces those
transitions, and restyles the zone announcements the game already makes.
Inspired by Nekres' Regions of Tyria for Guild Wars 2.

Repo: https://github.com/Yesanith/RegionsOfXIV
Release: https://github.com/Yesanith/RegionsOfXIV/releases/tag/v0.1.1.0
Licence: AGPL-3.0-or-later, matching Dalamud's
Notes for review

It hides some native UI, deliberately. The plugin replaces the game's own
location text rather than drawing a second notice on top of it, so by default it
suppresses _AreaText and the loading-screen title (_LocationTitle /
_LocationTitleShort) and draws its own in their place. Both are a single
checkbox away from coming back, and switching either off restores the game's
version immediately without a reload. The suppression is visibility-only
AtkUnitBase.IsVisible with nothing hooked or written into the addons, and
everything is restored on dispose.

It suspends automatic UI hide during loading screens. Entering a duty runs
the transition in two phases, and the second trips Dalamud's automatic hide,
which took the loading-screen notification with it partway through. The guard is
scoped to loading screens only (BetweenAreas, plus the NowLoading addon,
since the condition flag is not set across both phases) and is released the
instant one ends, so cutscenes and gpose still hide the UI normally. It is also
released on dispose so a stale suspension cannot outlive the plugin.

No combat elements. Nothing is read from or written to combat state. The only
condition flags consulted are for suppression cutscene, PvP, gpose, combat and
duty and PvP is suppressed unconditionally.

No game data is redistributed. Place names are read at runtime from the
player's own client via Lumina, so they are correct in every client language. The
one bundled asset is a fan-made Eorzean alphabet font used by the reveal effect;
its provenance and licensing are set out in
[NOTICE](https://github.com/Yesanith/RegionsOfXIV/blob/master/NOTICE), it is
expressly not covered by the project's AGPL, and the effect degrades to a plain
fade without it.

Settings use the Dalamud Windowing API, the version is fixed rather than
timestamp-derived, and the suppression rules are covered by unit tests
(dotnet test, no game or Dalamud install needed).

AI usage disclosure

Level: Assist
I built the foundation of this plugin myself and found the
addons it depends on (_AreaText, _LocationTitle, _LocationTitleShort,
NowLoading) through my own in-game testing. AI was used on demand for specific
tasks: understanding Dalamud's documentation and API surface while implementing
against it, and speeding up parts of the release build process. It did not drive
the design or architecture of the plugin. I understand, tested, and can explain
every part of this submission.


nathanctech:
bleatbot, approve

Yesanith:
Updated to v0.1.1.0

Fixes two cutscene bugs found testing in a dungeon: a notification could freeze
and then resume mid-animation after the cutscene ended, and an arrival that
happened during a cutscene was lost entirely.

Uses `DisableCutsceneUiHide` so a notification already on screen can finish.
Nothing new is started during a cutscene, so it adds nothing on screen.

nathanctech:
Make sure to update your manifest with the new commit hash.

Yesanith:
Manifest has been updated.

nathanctech:
Reading through the code it seems sane, but I must ask: was this really at "Assist" level? It has the hallmarks of mostly AI-written code with its verbosity and need to explain itself with every function of the code. It's important to disclose actual AI use in order to ensure future development won't be hampered by lack of knowledge when something the AI doesn't know how to solve comes up. I also see it's co-authored most of the commits in the repository.

Yesanith:
That's a fair concern. I built the base myself, then leaned on Claude quite a bit to speed things up. I work full-time in retail, so I often remote into my PC during breaks, have Claude implement specific changes, then review, test, and tweak everything when I'm back home. If it's something simple, I'll also review it during my breaks. That's why there are so many co-authored commits.

I don't blindly merge AI output though. I make sure I understand what it's doing and can maintain it myself before it goes into the repo. It also helps me get better at C#, not gonna lie. That's probably why there are a lot of comments all over the place. I'll also go through them and remove the obvious ones while keeping the ones that are actually useful for me.

PR #9066, opened 2026-07-27
[Testing] Karaoke v0.0.3.0

RajahOmen:
**Added**
 - More lyric fonts
   - Some should be dyslexic friendly
 - Import your own font file

### DIFF
--- a/testing/live/Karaoke/manifest.toml
+++ b/testing/live/Karaoke/manifest.toml
@@ -3,8 +3,9 @@ repository = "https://github.com/RajahOmen/Karaoke.git"
 commit = "35989ca65309e2acd8abe20a83d8957e8c686210"
 owners = ["RajahOmen"]
 project_path = "Karaoke"
-changelog = """
+changelog = """\
 **Added**
  - More lyric fonts
    - Some should be dyslexic friendly
  - Import your own font file
+"""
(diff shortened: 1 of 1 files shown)


nathanctech:
Hi, before I merge this, could you please provide the full list of fonts you're adding and the associated licensing, if you can find it? Fonts are a bit tricky as some may not permit embedding or distribution despite being "free", so just need to be on the safe side to be compliant.

RajahOmen:
Here's the font's I've added and what I've found about licensing (and the source linked). If I've made an error let me know and I'll fix it, I'm not that well versed in licensing.
- [Atkinson Hyperlegible, OFL-1.1 license](https://fonts.google.com/specimen/Atkinson+Hyperlegible/license?preview.script=Latn)
- [Comic Relief, OFL-1.1 license](https://github.com/loudifier/Comic-Relief?tab=OFL-1.1-1-ov-file)
- [Open Dyslexic 3, SIL-OFL license](https://antijingoist.itch.io/opendyslexic)
- [VCR OSD Mono](https://www.dafont.com/font-comment.php?file=vcr_osd_mono), I wasn't able to find the license exactly, but where I got it marks it as "100% free", and the author uploading it clarifying in the comments section that its free for commercial purposes. The download also did not include a readme/license file.

I don't have specific plans on adding any more in the future unless there's one particular one being requested a lot since users can import their own fonts. But if I did I would do the same check for licenses and make sure to post it in the D17 PR description.

nathanctech:
Thanks, it doesn't have to be distributed with your plugin (no need to update), but please include a text file in your repo with that list just to be safe. It looks like you're fine as long as the fact you're using it is called out somewhere.

nathanctech (approved):

PR #9072, opened 2026-07-28
Faux Hollows Solver API 15

Glorou:
Updated Faux Hollows Solver for API 15


nathanctech:
Hi, please remove the `<PackageReference Update="DalamudPackager" Version="15.0.0" />` from your csproj. This is newly enforced for API 15 since it may break with future releases.

nathanctech (approved):

PR #9109, opened 2026-08-01
image size adjusted and changed

rail2025:
i just fixed the image previews that weren't showing up in the dalamud installer:
| ERR | Plugin image1 for BeastieBuddy at queue was larger than the maximum allowed resolution (432x399 > 730x380).
| ERR | Plugin image2 for BeastieBuddy at queue was larger than the maximum allowed resolution (424x406 > 730x380).
also changed one. 
There aren't any code changes. If this doesn't work then I'll just add it to next code patch.


rail2025:
fission mailed

PR #9126, opened 2026-08-03
[Testing] Housingway v1.0.2.1

Abyeon:
## Housingway
A tweak collection plugin for housing. Custom skyboxes, helper tools, and general adjustments for all your housing needs!

## Current Tweaks:
* Disable Camera Collision
* Disable Cast Shadows
* Disable SSAO
* Display Pop Range
* Door Command
* Furniture Info
* Highlight Phased Objects
* Indoor Sun
* Model Adjustments
* No Positional Orchestrion
* Override Interior Lighting
* Skybox

More information can be found on the [plugin repo](https://github.com/Abyeon/Housingway)


nathanctech:
Hello,

PAC has some concerns with this one as housing plugins tend to already be a grey area for main repo. The biggest one is the feature to disable the "box" for voidbuilding - since this is against ToS technically, we cannot allow this feature into the main repo. Anything that could help the user build outside of SE's intended systems would be a no-go for main repo as well, but I haven't investigated all the features in detail yet. Just something to keep in mind during review.

Abyeon:
Thank you for reviewing the plugin! I would like to say- I don't see how disabling the visibility of a model assists anyone with void building. It's not like I'm providing any tools for placing furniture. If you could elaborate I would appreciate that. I also am not aware of void building being explicitly against ToS- as this is entirely possible without third party tools, could you point me to where SE has taken an official stance on that? I definitely could have missed it.

nathanctech:
> Thank you for reviewing the plugin! I would like to say- I don't see how disabling the visibility of a model assists anyone with void building. It's not like I'm providing any tools for placing furniture. If you could elaborate I would appreciate that. I also am not aware of void building being explicitly against ToS- as this is entirely possible without third party tools, could you point me to where SE has taken an official stance on that? I definitely could have missed it.

From what I understand, it can fall under the use of glitches but I know in interviews Yoshi-P has mentioned that it's more or less a "don't ask don't tell" type of situation with clipping furniture and such to make a lot of designs. Voidbuilding (making an area larger than a typical house can usually support) might be a different matter. I only mentioned it since that particular tweak was geared towards that purpose....probably needs a bit further discussion because a lot of these tweaks are pretty harmless otherwise.

Abyeon:
I understand why voidbuilding may specifically cause issues, but again I do not actually provide tools to assist with that from my perspective. Is this something that all of PAC has weighed in on? If not I definitely would love to have more perspectives so I can understand why exactly this may be problematic.

nathanctech:
Not yet, hence why I'm not saying the features have to go just yet :-) just one thing I noticed that could be a concern.

Abyeon:
Ah! No worries, thanks for all the hard work Era <3

Abyeon:
@nathanctech Hey! Just checking in, has there been any further discussion with PAC about the issues you had? If not, would it be worth making a post in #plugin-dev?

nathanctech:
I will ask, sometimes it takes a bit of prodding.

PR #9143, opened 2026-08-05
[Testing] ARealmReorganized

Nepharyas:
First submission. glamour dresser cleanup helper: scans the dresser and shows what can move into the 7.5 armoire, finds partial and complete sets, spots duplicates across the dresser, armoire, bags, armoury, saddlebag and retainers, and highlights the matching items in-game across the inventory windows so users don't have to hunt for them.

Repo is AGPL-3.0 with unit tests on the detection logic.

AI usage disclosure:
GenAI (Claude Code) was used to help me navigate through ClientStructs more quickly, write some tests, and with C# shenanigans (i'm just a lil' python dev...)


reiichi001:
bleatbot, approve

reiichi001:
Your github repository must be public.

Nepharyas:
whoops sorry, thought I put it in public! made the change now, thanks for telling me

nathanctech:
Sorry for the delay, here's some code review notes:

[Non-blocking] InventoryReader.cs: The ReadAll() reads cached data from the client - however, saddlebag is a special case so if you run a scan when it's not available (inside instances, for example) it will return an empty set and cause incomplete data. I don't think it'd be a common case, but something to keep note of.

[Non-blocking] Running an on-demand scan is on the framework thread and could cause hitching, but needs testing.

[Non-blocking] I would look at using ImRaii where possible (like for the begin/end child, etc) which has better safeguards if the drawing is stopped somehow.

[Non-blocking] The logging seems a bit over-engineered, I would just stick with the provided IPluginLog service as the Dalamud-provided service uses Serilog which handles all the buffering and other tasks for you.

I may have more feedback after it can be properly tested, but these would be a good focus if you wanted to address any of them.

Nepharyas:
Thanks a bunch for your feedback!
I will definitely look into your suggestions, with logging as priority, I didn't look Dalamud's own IPluginLog logging enough, thanks again! Guess I got a bit overzealous xD

Nepharyas:
Hi again, back with some updates, thanks again for your review!

Logging: dropped the previous logging system and the settings window that displayed it, everything goes through IPluginLog now.  It was this way because of a feature idea I had which I ended up removing, and I didn't catch that it rendered my logging system obsolete

Saddlebag: every container gets an IsLoaded check now, and the scan reports back whether the saddlebag was readable so the two tabs that use its contents can say when it wasn't. Picked up PremiumSaddleBag1/2 while in there, we weren't reading those at all...

Scan on the framework thread: measured 18ms on the first scan of a session, 1ms on every one after, debug build with 757 dresser items and a full armoury. I think the gap between the first and the rest is lumina paging the sheets in on first touch rather than the scan itself. Left the timing in at debug level so the number is there next time it comes up. also folded two full sweeps of the set sheet into one pass, and stopped the scan from resolving item names that the draw path already resolves lazily.

ImRaii: done, every begin/end pair is gone. child windows, the tab bar and its items, style colours and text wrap.

Another bug I found and fixed was a 10 second cache window and multiple json writes on the framework thread. Now the dresser and armoire caches only write on a real change now. verified it sits untouched through 25 minutes of play, then
writes once when an item moves.

nathanctech:
bleatbot, rebuild

nathanctech:

nathanctech (approved):

Nepharyas:
Great, thank you very much! My discord handle is nepharyas as well

PR #9170, opened 2026-08-08
Allagan Market 1.4.0.2

Critical-Impact:
### Changed
- Update sig for 7.55


nebel:
I think `RetainerMarketService.cs` needs to be modified to resolve the ambiguity of `AgentRetainer` when building against latest Dalamud version.

nebel (approved):

PR #9173, opened 2026-08-08
[New Plugin] PfExplorer

nnillni:
PF Explorer allows the players to visualize Party Finder listings captured by other players (crowdsourced) as you play and shares them to pfexplorer.com, so you can search Party Finder across every data center instead of just your own, even inside instances which you usually can't.

  Features:
  - Cross-DC search — see and filter listings from every data center, not just the one you're on
  - Works even while you're inside an instance, so you can keep browsing results without leaving duty
  - Click a listing to open the join window directly
  - One-click travel prompt to other data centers via Lifestream
  - Notifications for new results, party size changes, and listings closing — each with its own configurable chat color and clickable chat links
  - Full results window or a compact minimal view
  - Free-text search box, plus filters by job, item level, data center, category, freshness, and Blue Mage
  - Optionally aggregates xivpf.com results alongside listings captured by pfexplorer.com

Some more info at [README.md](https://github.com/nnillni/pfexplorer-client/blob/main/README.md)

  ## AI Usage Disclosure

Per the AI Usage Policy (https://dalamud.dev/plugin-publishing/ai-policy/), this project's disclosure level is **Copilot** — see [AI-DECLARATION.md](https://github.com/nnillni/pfexplorer-client/blob/main/AI-DECLARATION.md) in the plugin repo. 

Developed with Claude Code throughout; all design, architecture, and merge decisions were made by a me and tested every change before it shipped.


reiichi001:
bleatbot, approve

nathanctech:
[Blocking]

You cannot poll the server for PF requests as it is very detectable by SE. You must remove this feature in order for the plugin to be considered further.

nnillni:
@nathanctech
 The feature was force-disabled and couldn't be activated because I was advised in discord to avoid it.
 
 I'll remove that disabled checkbox and code related to it, so only the code that passively captures from when the player opens the party finder works, like the outdated Remote Party Finder Uploader did

image

nathanctech:
Thank you for removing the code. As part of our policy, we don't allow these features even if disabled by default. Your plugin is in our review queue to be looked at.

Infiziert90:
Gave your code a quick look and found some issues that should be resolved or may result in users having a bad time with the plugin / crashing to desktop

- Please use ImRaii, there is a multitude of unchecked states and a lot of push/Pop code that can corrupt other plugins
- Respect the users choosen font size and global scale, scale hardcoded values with `* ImGuiHelpers.GlobalScale`
- Build your own font with a smaller/bigger size via UiBuilder, using ImGuis `SetWindowFontSize` is a very simple image scaling which will result in it looking horrible with bigger font sizes
- I'm pretty sure that this plugin does not follow the 1:1 rule, as I have seen some code parts that should be a 2:1, opening PF + joining a listing
- What is the BlueMage AOZ code for?
- Security through some algorithm that is baked into the image is not a good idea, with everything being open source it brings nothing
- Custom repo plugins, especially a bot like that one, shouldn't be mentioned anywhere user facing

nnillni:
Diff [v0.2.0...v0.3.0](https://github.com/nnillni/pfexplorer-client/compare/v0.2.0...v0.3.0)

> * Please use ImRaii, there is a multitude of unchecked states and a lot of push/Pop code that can corrupt other plugins

Done for colors/tables/tabs/combos and such

> * Respect the users choosen font size and global scale, scale hardcoded values with `* ImGuiHelpers.GlobalScale`
> * Build your own font with a smaller/bigger size via UiBuilder, using ImGuis `SetWindowFontSize` is a very simple image scaling which will result in it looking horrible with bigger font sizes

Both should be done

> * I'm pretty sure that this plugin does not follow the 1:1 rule, as I have seen some code parts that should be a 2:1, opening PF + joining a listing

Yeah, didn't know about this rule. I have reworked the code to be 1 action 1 command. Technically I can open directly the detailed window to join the PF, but I kind of need the players to also open the party finder window so the results can be captured and sent for other players (otherwise it would canibalize itself). So for now you'll have to double-click I guess.

> * What is the BlueMage AOZ code for?

This is just a small feature I added to open the blue magic spellbook (same as /bluespellbook or clicking the menu option) in the blue mage tab.

image

> * Security through some algorithm that is baked into the image is not a good idea, with everything being open source it brings nothing

Yeah it's security by obscurity and adds not much, was just a bit bored, changed that.

> * Custom repo plugins, especially a bot like that one, shouldn't be mentioned anywhere user facing

Removed specific references let me know if its fine now

nathanctech:
Just browsing the diff, I would just remove the "Travel" button as there's no method to support automating this and we would rather not see it in the code. https://github.com/nnillni/pfexplorer-client/compare/v0.2.0...v0.3.0#diff-7fda8148be954d76580a4c640f46f17415a4da0f58f256d7eceed6868523c2d1R296

nnillni:
Removed the travel bridge functionality altogether + some QoL and small fixes

Diff [v0.3.0...v0.3.1](https://github.com/nnillni/pfexplorer-client/compare/v0.3.0...v0.3.1)

nathanctech:
Looking through code review again:

[Blocking]

Hashing the ContentId with a fixed string is no security at all. Since this is a party finder plugin you do need some sort of identifier to see who the recruiter is, but a name should be sufficient. This data should be encrypted at rest on the server and scrubbed as soon as it's no longer required.

[Blocking]

As all code is licensed under AGPL, you must also provide the server-side code in a public place for review. This can be in a separate repository or as a separate project in your plugin's repository.

Edit to add: I also have concerns about the actual security implementation as it's easily defeatable with public code - I would highly advise looking at a better method to combat possible abuse (fake listings, etc).

nnillni:
closing for now, might re-add in a future

PR #9201, opened 2026-08-15
[Testing] VanillaPlus

MidoriKami:
nofranz

Some experimental code changes

### DIFF
--- a/testing/live/VanillaPlus/manifest.toml
+++ b/testing/live/VanillaPlus/manifest.toml
@@ -1,5 +1,5 @@
 [plugin]
 repository = "https://github.com/MidoriKami/VanillaPlus.git"
-commit = "72536582879a1d6cec7ef713f5c16c4b3b3ea989"
+commit = "375ebf3059ba9245fb3e87523c88e5389fa26d71"
 owners = [ "MidoriKami", "Haselnussbomber",]
 project_path = "VanillaPlus"


MidoriKami:
Diff from stable -> testing [REAL DIFF](https://github.com/MidoriKami/VanillaPlus/compare/958c461556997660f1e1feb75d52b14785cc981b..375ebf3059ba9245fb3e87523c88e5389fa26d71)

No functional changes, just some style/system changes, but these were done maybe a bit hastily so I might have missed or broken something.

Theoretically these changes will make loading modules one to three frames slower to load (still all in parallel) but might make it easier to debug an issue if one occurs.

MidoriKami:
Summary of changes:

In the previous version I tried adding some helpers to make it simpler to await various disposes, but I think those helpers may have made things worse, so I have removed them and added new helpers!

Most notably a IAsyncDisposable.DisposeAsyncSafely, that will do the null check internally so I don't have to write `await (Thing?.DisposeAsync() ?? Task.CompletedTask);` a million times

I also added EnableAsync/DisableAsync/DisposeAsync to a bunch of KTK classes so they can be `await thing.DisposeAsyncSafely()` easily.

And finally, all hooking is now done on the main thread via `Hook<T>.EnableAsync` and `Hook<T>.DisposeAsync` this wasn't really necessary as we don't really have problems hooking off the main thread, but adds an extra layer of safety around hooking.

nathanctech (approved):

PR #9211, opened 2026-08-16
[Testing] Add ItsUp

juan-medina:
It's Up shows a reminder icon for an ability once it's off cooldown, but only for abilities the player has actually used - nothing is shown pre-emptively. 

It only surfaces the player's own recast/charge state, already visible via the hotbar's recast swipe; no automation, no server interaction, no data collection.

AI usage disclosure: Assist - AI used on demand for specific tasks, e.g. split this function, but all changes reviewed, validated and tested by me


nathanctech:
bleatbot, approve

nathanctech:
Did some testing and it seems to not work correctly. As DNC, I used standard step which it correctly counted down the 5 seconds, but then disappeared. En Avant did not show at all. I have the following settings:

image

juan-medina:
I tested both on DNC and they do work, but the trigger logic is very specific:

The plugin only alerts when an ability transitions from unusable back to usable after you've used it in combat. It answers "can I press this again?", not "is this available?"

- Charge abilities: It only triggers when you go from 0 charges to 1. If an ability has 3 charges and you sit on 1 or 2, it will never ping because you were already able to press it.

- Why Closed Position didn't show: You have to be in combat and actually use it first so it enters cooldown. It won't pop up for unused skills sitting off CD so your screen doesn't get spammed when a fight starts.

- Better test skills: Things like Closed Position are a bit weird for this—it shines on rotational stuff you don't want to drift like Devilment, Standard Step, or Technical Step.

Also updated the manifest to the latest test build with some recent fixes just in case.

Thanks again!

nathanctech:
I see, I think I misunderstood the "until pressed" aspect, it makes more sense now. I'll do a quick code review when I can this evening and it should be good to go after that.

nathanctech:

nathanctech (approved):

juan-medina:
Yes I'm the dalamud discord:

 id: 205346961044406272

> Congratulations on your first plugin added to the main Dalamud plugin repository!
> 
> As a first-time contributor, if you're in the `XIVLauncher & Dalamud` Discord server and would like to get access to the private developer channels for plugin feedback and others, please let us know your Discord handle and we'll grant you the role.

PR #369, opened 2022-09-03
[Testing] DailyDuty v3.0.1.7

MidoriKami:
More siginificant internal changes laying groundwork to implement more advanced modules.
*Look forwards to what's coming*

This set of changes targets GoldSaucer modules, please make sure to watch the status tab closely and report any odd behavior.

`Wondrous Tails` Fixed clickable links not working
`System` Improved saving and loading, may fix weird rare migration bug

PR #9216, opened 2026-08-17
Discord Rich Presence 2.1.0.0

Bronya-Rand:
# What's New?
- Rebuilt from the ground up the Wine bridge for non-`AF_UNIX` Wine/Proton builds on Linux and macOS using A TCP RPC bridge.

## Note
This is **not** the same bridge that once existed and currently can be pushed despite the pending PR in XIVLauncher.Core as a hostable binary is avail on the NuGet repo I made that XOM and XIVLauncher-RB uses.

### DIFF
--- a/stable/Dalamud.RichPresence/manifest.toml
+++ b/stable/Dalamud.RichPresence/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Bronya-Rand/Dalamud.RichPresence.git"
-commit = "08db5a31a96390abe3b59b6fe66e9be7b9cfb5aa"
+commit = "734b29685917f30663a0b787a6d6b7dc4df36234"
 owners = [
     "Bronya-Rand"
 ]
--- a/stable/Dalamud.RichPresence/manifest.toml
+++ b/stable/Dalamud.RichPresence/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Bronya-Rand/Dalamud.RichPresence.git"
-commit = "734b29685917f30663a0b787a6d6b7dc4df36234"
+commit = "173370383f7d066a4276050ee8ffabe2ff38dd9c"
 owners = [
     "Bronya-Rand"
 ]
--- a/stable/Dalamud.RichPresence/manifest.toml
+++ b/stable/Dalamud.RichPresence/manifest.toml
@@ -1,6 +1,6 @@
 [plugin]
 repository = "https://github.com/Bronya-Rand/Dalamud.RichPresence.git"
-commit = "ba4e3371d42496fc235548b21c33ef80f004ca7d"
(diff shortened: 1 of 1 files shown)


Bronya-Rand:
Forwarding this from Discord as a note:

I get this update is a bit of a hot topic given it requires users to run a binary file to run Discord RPC between XIV on Wine and Discord but there exists a reason why this is the option I had to go with to get feature parity between the Socket code (which uses `AF_UNIX` which is experimental on Wine-Staging) and other Wine/Proton clients + macOS for those that do not have this protocol:

1. The problem is literally whether the version of Wine/Proton supports AF_UNIX or not. There is no third option regarding running it in the Wine prefix that's maintainable for me to write code for.
2. The only reason why the old WineRPCBridge worked is because it uses 32-bit syscalls which bypasses Wine to call Linux directly but that's fragile and looking at the Git code for it, it is archaic to maintain in the future.
3. I already tried to reduce the footprint of errors regarding RPC on Linux/macOS. I provided the README, updated the forum, added a setup guide, already set the files to be 755 by default.
4. Wine cannot auto-start Linux/macOS binaries. I tried to make it automated but no matter what I tried, it just refused to work with every command I tossed at it.

XIVDeck and current RPC break on macOS and will break on older Wine/Proton from both XIVLauncher and XIVLauncher-RB. There had to be a fallback between sockets somewhere and if not every Wine/Proton is going to ship with AF_UNIX since it is still on Wine staging or the implementation is broken, then the only alternative way that didn't risk triggering stuff on Windows, doesn't rely on sys-calling the kernel which is more ASM than C, and worked for every client like the new socket code was literally to make the plugin fetch data from a host binary that bypasses Wine emulating WinSock to connect to Discord and pass in data to and from Discord to XIV via localhost TCP. If there was another way to do it, I would have done it but from what research I put into it it is literally "RPC or no RPC" for many XIV users on Linux/macOS.

I know that having the user open a terminal to start up a bridge can be risky for new users or anyone really, but after putting every detail into seeing if anyone else solved the problem, the shipped binary + Terminal to execute it is, unfortunately, the only way to get RPC working with every Discord client on Linux, whether its Flatpak/Snap/Native or Vesktop. That or have XLCore ship the bridge I made and it starts/exits when the game boots or quits.

Bronya-Rand:
After discussion with PAC 8 hours ago and some finding, the binary requirement has been dropped. However PR is blocked until XIVLauncher.Core PR [#369](https://github.com/goatcorp/XIVLauncher.Core/pull/369) is merged and is released as a XIVLauncher update.

Bronya-Rand:
To update PAC on these next updates: The latest commits add a envar that's needed for macOS. Also, given Goat and Blooym are debating the talk on the bridge on XIVLauncher.Core still, I made external binaries again that are hosted on a separate repo with source code that the user can download (similar to XIVDeck).

The PR is feature complete now so it can be merged even if XIVLauncher.Core PR is still being discussed ATM.

nathanctech (approved):

PR #9234, opened 2026-08-22
VanillaPlus

MidoriKami:
Fix failing to load some modules due to ClientStructs changes.


MidoriKami:
context: when using Enum.GetValues<T>.ToDictionary, if the enum has any Obsolete values, this causes a duplicate key exception. This update will filter out Obsoleted entries.

nathanctech (approved):

PR #9247, opened 2026-08-25
Add Hot Potato Game Manager

MarcusBullock22:
Adds Hot Potato Game Manager to the Dalamud testing track.

Hot Potato Game Manager is a party-based dice game manager that tracks participating players, randomized Hot Potato numbers, rounds, turn order, eliminations, and party chat announcements.

The plugin can be opened with `/hotpotato`, with players rolling using `/dice party`.

### AI Usage Disclosure

AI tools were used as a development assistant during this project for portions of code generation, debugging, refactoring, and documentation. I defined the plugin's requirements and game behavior, reviewed and modified the generated code, and personally tested the plugin and its functionality. I understand and maintain the submitted code.


nathanctech:
bleatbot, approve

nathanctech:
[Blocking]

It appears your plugin automatically sends messages to chat. This would absolutely be detectable by SE (or could be reported by players) and we will not permit chat automation. This is usually why "game manager" plugins aren't really present as automating chat is how they usually work.

Infiziert90:
Just gave it a quick look and as expected it uses Regex to parse the numbers, which does not work in other languages...

DeathRollHelper has for 2 years already code that grabs random and dice numbers with the games own formatting functions, so please do the same instead of using Regex

MarcusBullock22:
Alrighty, I went through the feedback and pushed up the changes to address everything that was called out.

- Removed the automatic party chat messages. The plugin will still queue up the announcements, but the user now has to click Send to Party before anything is actually sent.
- Reworked the dice detection to get rid of the regex/chat parsing. I used the DeathRoll implementation as a reference and am now pulling the player name, roll, and roll limit directly instead of trying to parse the displayed chat message. This should also take care of the localization issue that was pointed out.
- Cleaned up the old IChatGui/chat parsing code that isn't needed anymore.
- Ran through the changes in-game and tested game start, normal rolls, Hot Potato hits/eliminations, winner handling, sending queued announcements, and skipping announcements.
- Also fixed the screenshot sizing from the failed check and updated the manifest to point to the new plugin commit.

Everything is building clean and the updated flow is working in-game on my end. Appreciate you guys taking the time to look through it and point me in the right direction!

nathanctech:
I will need to discuss further with the rest of PAC, but I still have concerns with the party chat usage since it will definitely come off as "bot-like" to other players even though it takes place in party chat. Minigames played over text in-game are very dicey because it runs the risk of the user being reported - and the GM who reviews the logs may not know or care the user is manually submitting each message.

MarcusBullock22:
> I will need to discuss further with the rest of PAC, but I still have concerns with the party chat usage since it will definitely come off as "bot-like" to other players even though it takes place in party chat. Minigames played over text in-game are very dicey because it runs the risk of the user being reported - and the GM who reviews the logs may not know or care the user is manually submitting each message.

No worries if not, it was a little plugin my wife had me whip up.

PR #9248, opened 2026-08-25
[Testing] MountTilt v0.0.0.1

RokasKil:
A simple plugin that allows you to apply a multiplier to your mount tilt angle and speed when in motion.


nathanctech:
Only real feedback I have is: https://github.com/RokasKil/MountTilt/blob/master/MountTilt/Services/TiltService.cs#L21

I don't know if it happens or not, but I would put a null check here or use `?.Enable()` since it could crash if your hook can't be made for some reason.

RokasKil:
> Only real feedback I have is: https://github.com/RokasKil/MountTilt/blob/master/MountTilt/Services/TiltService.cs#L21
> 
> I don't know if it happens or not, but I would put a null check here or use `?.Enable()` since it could crash if your hook can't be made for some reason.

Don't think it can happen since `GameInteropProvider` throws if it fails to make a hook, but either way if the returned hook somehow did end up null a `NullReferenceException` would be thrown and it would just fail to load which is the desirable outcome since it's an unrecoverable error, a null check would leave the plugin loaded but nonfunctional.

nathanctech:
bleatbot, rebuild

nathanctech:
Thanks for your contribution and your patience. Your plugin will be merged in shortly!

nathanctech (approved):

PR #9258, opened 2026-08-26
Add HidePlayers

tyim702:
Adds HidePlayers to the testing/live plugin repository.

HidePlayers is a client-side player visibility plugin that hides other player characters while optionally keeping party members and friends visible.

### Features
- Hides other player characters and their nameplates
- Keeps the local player visible
- Optionally keeps party members visible
- Optionally keeps friends visible
- Automatically handles players entering and leaving the area
- Restores hidden players when the plugin is disabled
- Saves configuration between sessions

This is the initial release of HidePlayers (v0.1.0).


rail2025:
https://github.com/SheepGoMeh/VisibilityPlugin

If you weren't aware of a pre-existing plugin.

nathanctech:
While we do generally allow plugins with similar functionality, this one appears to do the exact same thing the Visibility plugin does, as pointed out. Is there something this plugin does that's missing from the already published one?

nathanctech:
Hi @tyim702,

Your submission closely overlaps with functionality of an existing plugin in the main repo and since I have not heard from you in about a week I will close this submission for now. Should you wish to submit, please reply to this PR. Thanks!

PR #9280, opened 2026-08-31
[Testing] BardicMemory v0.0.1.0

OOFGamesss:
# Bardic Memory

Guess the tune from a sliver of FFXIV's orchestrion library.

Bardic Memory picks a random orchestrion roll, plays you half a second of its real opening and asks you to name it. Name it on that first sliver and the round is worth six points. Stretch the clip through 1, 2, 4, 8 and 15 seconds and a point comes off each time you do, but a wrong guess costs nothing, so you can keep naming songs at the length you are on.

Search every named roll as you type with titles ranked first, then descriptions, then the places and duties a track plays in. Filter the pool by orchestrion category from locales and dungeons through to raids, trials and seasonal music and leave out anything you have already solved. When the round ends a reveal names the track, gives you its description and the zones it plays in and offers the full piece before the next song. Every finished round lands on the history tab with running totals, streaks and solve rate.

Inspiration for Bardic Memory's functionality and UI: 
- https://songspot.net/

Credits to the Orchestrion Plugin creators for code and orchestra mapping.


OOFGamesss:
bleatbot, rebuild

PR #9299, opened 2026-09-05
XIVJitterFix Remove duplicated outdated entry from stable folder

MikeZTM:
This plugin was once a stable plugin and with my revival it has a outdated entry in the stable folder.
Removing it to turn it into a test version plugin.


reiichi001:
Dalamud doesn't show plugins that are more than 2 API versions old. Our build system also requires...a build. There's no reason to remove the stable folder. Users will not see it unless they've already opted into plugin testing builds.

MikeZTM:
I just figure that out with in-game plugin installer. Thus closing this one. Thanks for the notice!

PR #9308, opened 2026-09-06
ChillFrames

MidoriKami:
Fix idle limiter not being disabled correctly during loading zones.


MidoriKami:
It thinks the window file is completely changed because I also switched it from tabs to spaces, semantic diff is a much better view for that file.

reiichi001 (approved):

PR #9354, opened 2026-09-12
ItsUp v0.7.0.0

juan-medina:
Show hotbar keybinds on ability icons.
Role abilities are now tracked per job.
Added an option to only show while inside duties.
Added an option to toggle the marching ants border animation.
Improve config window.
Added /itsup preview and /itsup reset slash commands.


nebel (approved):

nebel:
@juan-medina Please note that slash commands being listed in the new description means they are shown twice in the plugin installer.

PR #9374, opened 2026-09-16
[Testing] Keybinds Plus 0.1.0.0

Bronya-Rand:
Keybinds Plus is a plugin that allows you to bind _almost_ any key on your keyboard to either a XIV keybind or a supported Dalamud plugin.

Keybinds Plus allows you to bind any key, including those that are not supported by the game (F13-F24) to almost any game function (Shortcuts, Chat, Targeting, Hotbars) as well as any plugin that implements KeybindsPlus' IPC. You can also bind a key to a custom command (like `/wlmark`) or macro just like how the game does.

Supported Keybinds:
- Custom Macros and Commands
- Supported Plugins
- Targeting*
- Shortcuts
- Chat*
- Hotbars

  > \* - Some features may be grayed out due to complexities or unavailability. Not all base game keybinds are in KeybindsPlus but may appear as updates progress.

This plugin is a early state and quite ambitious of me to do from prior plugins I've made. Why did I decide to write this? Simply put I wanted to use my Razer Miyabi keyboard macros and originally wanted to use F13 but game said no so I said `Fine. I'll do it myself.`  Also, past conversations. Thanks Miyabi :)

## AI Declaration
Code Usage: Assist
> Claude/Gemini was used only to inquire about what ClientStructs to use given my prior research to reading said ClientStructs aside from the people who suggested/pointed me to the right CS in #plugin-dev and reading code from other plugins that were a basis on Better Keybinds like Vanilla Plus, MouseLock, Chat 2, a bit of DelvUI and Better Targeting System in a way I could best understand it. All other code was written by me, borrowed from some Github searching for things and Google-fu.

Image Usage: None
> The icon (the M5 key button with the "No. 564" Miyabi identifier) was made by me in Photoshop with some SVG modifications to the "No. 564" image itself.


nebel:
Hi @Bronya-Rand 

Thanks for the submission. I did some testing of your plugin and one issue I wanted to bring up is that it is very aggressive in terms of the amount of logging it does. Every enabled bind will log one line per (unrelated) keypress at DEBUG level, and holding many keys also logs once per frame.

For development this is fine but I think it's better to disable some of this (or set the level to VERBOSE) in release builds as it could contribute to troubleshooting logs being more difficult to read.

Bronya-Rand:
> Thanks for the submission. I did some testing of your plugin and one issue I wanted to bring up is that it is very aggressive in terms of the amount of logging it does. Every enabled bind will log one line per (unrelated) keypress at DEBUG level, and holding many keys also logs once per frame.

Noted. Will remove and ask a PAC member to push. Thanks!

